PR #11, opened 2018-08-14
Rename the `topic` attribute to `name`

cescoffier:
Related to #8

Just renamed the `@Incoming` and `@Outgoing` `topic` attribute to `name`.


cescoffier:
Right, my bad, will update.

PR #16, opened 2019-02-01
Initial proposal for the Core TCK

cescoffier:
This PR

* updates the current code base to Reactive Streams Operators 1.0
* Fix javadoc
* define the core TCK (signature, scope...)

### DIFF
--- a/core-tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/SimpleIncomingTest.java
+++ b/core-tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/SimpleIncomingTest.java
@@ -31,6 +31,7 @@ import javax.inject.Inject;
 import java.util.ServiceLoader;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
 
 @RunWith(Arquillian.class)
 public class SimpleIncomingTest {
@@ -50,7 +51,8 @@ public class SimpleIncomingTest {
 
   @Test
   public void testReceptionWithValues() {
-    assertThat(simple.getValues()).containsExactlyElementsOf(StringSource.VALUES);
+      await().until(() -> simple.getValues().size() == StringSource.VALUES.size());
+      assertThat(simple.getValues()).containsExactlyElementsOf(StringSource.VALUES);
   }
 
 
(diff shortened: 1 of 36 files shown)


core-tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/SimpleIncomingTest.java
```diff
@@ -0,0 +1,57 @@
...
+    ServiceLoader.load(ArchiveExtender.class).iterator().forEachRemaining(ext -> ext.extend(archive));
+    return archive;
+  }
+
+  @Inject
+  private SimpleIncomingBean simple;
+
+  @Test
+  public void testReceptionWithValues() {
+    assertThat(simple.getValues()).containsExactlyElementsOf(StringSource.VALUES);
```
jroper: Is there a race condition here? How do we know that the bean has finished consuming all the values when the test runs?

core-tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/scope/ApplicationScopeTest.java
```diff
@@ -0,0 +1,48 @@
...
+      .addClasses(ApplicationBeans.class);
+  }
+
+  @Inject
+  private ApplicationBeans bean;
+
+  @Test
+  public void verify() {
+    assertThat(bean.getList()).containsExactly(1);
+    assertThat(ApplicationBeans.getStaticList()).containsExactly(1);
```
jroper: Likewise, is there a race condition here?

core-tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/SimpleIncomingTest.java
cescoffier: Would be better with `await`.

cescoffier: Let's use `await` everywhere.

cescoffier: Fixed

core-tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/scope/ApplicationScopeTest.java
cescoffier: Fixed.

cescoffier:
@jroper I've updated the PR to add awaitility and the related statements to avoid the race conditions.. How should we proceed?

PR #22, opened 2019-03-26
Make channel name mandatory and connector support

cescoffier:
This PR is the result of the discussions from the _reactive hangout 2019-03-26_

* Make `@Incoming` and `@Outgoing` values mandatory
* Connector SPI
* Extend the TCK to enforce the connector SPI
* Extend the TCK to reject blank stream names
* (Adds dependency on MicroProfile Config)


api/src/main/java/org/eclipse/microprofile/reactive/messaging/Incoming.java
```diff
@@ -88,34 +90,8 @@
 
     /**
      * The value of the consumed stream.
```
hutchig: This should say "The name of the stream being consumed".  There are some other small changes I would like to put into the doc in text that is not part of this PR so I will do a small PR after this one is merged.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+import org.eclipse.microprofile.reactive.messaging.Message;
+import org.eclipse.microprofile.reactive.messaging.MessagingProvider;
+import org.eclipse.microprofile.reactive.streams.operators.PublisherBuilder;
+
+import java.util.NoSuchElementException;
+
+/**
+ * SPI used to implement a connector managing a source of messages for a specific <em>transport</em>. Typically, to
+ * handle the consumption of records from Kafka, the reactive messaging extension would need to implement a {@code bean}
+ * implementing this interface. This bean is called for every {@code stream} that need to be created for this specific
```
hutchig: "Typically..." -> "For example, to
 * handle the consumption of records from Kafka, the reactive messaging extension would need to implement a {@code bean}
 * implementing this interface that reads records from the identified Kafka source."

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.reactive.messaging.Message;
+import org.eclipse.microprofile.reactive.messaging.MessagingProvider;
+import org.eclipse.microprofile.reactive.streams.operators.PublisherBuilder;
+
+import java.util.NoSuchElementException;
+
+/**
+ * SPI used to implement a connector managing a source of messages for a specific <em>transport</em>. Typically, to
```
hutchig: "Typically," -> "For example,"

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
hutchig: "need" -> "needs"

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * handle the consumption of records from Kafka, the reactive messaging extension would need to implement a {@code bean}
+ * implementing this interface. This bean is called for every {@code stream} that need to be created for this specific
+ * <em>transport</em> (so Kafka in the example). These streams are connected to methods annotated with
```
hutchig: "in the example" -> "in this case"  (or perhaps "in this example")

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * MicroProfile Config. The following snippet gives an example for a hypothetical Kafka connector:
+ *
+ * <pre>
+ * mp.messaging.incoming.my-stream.type=i.e.m.reactive.messaging.impl.kafka.Kafka
+ * mp.messaging.incoming.my-stream.bootstrap-servers=localhost:9092
+ * mp.messaging.incoming.my-stream.topic=my-topic
+ * ...
+ * </pre>
+ * <p>
+ * This configuration keys are structured as follows: {@code mp.messaging.[incoming|outgoing].stream-name.attribute}.
```
hutchig: "This" -> "The"

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * {@code Incoming} annotation:
+ *
+ * <pre>
+ * &#64;Incoming("my-stream")
+ * public void consume(String s) {
+ *      // ...
+ * }
+ * </pre>
+ * <p>
+ * The set of attributes depend on the connector and transport layer (typically, bootstrap-servers is Kafka specific). The
```
hutchig: "typically" -> "for example,"

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * {@code type} attribute is mandatory and indicates the fully qualified name of the {@link MessagingProvider}
+ * implementation. It must match the class returned by the {@link #type()} method. This is how a reactive messaging
+ * implementation looks for the specific {@link IncomingConnectorFactory} required for a stream. In the previous
+ * configuration, the reactive extension implementation would need to find the {@link IncomingConnectorFactory} returning
+ * the {@code i.e.m.reactive.messaging.impl.kafka.Kafka} class as result to its {@link #type()} method to create the
+ * {@code my-stream} stream. Note that if the connector cannot be found, the deployment must be failed.
```
hutchig: Do you want to say " with a DeploymentException" ?

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
hutchig: "The portion of the key that precedes the {@code attribute} is acts as a property prefix
that has a common structure across all {@code streams}."
(to introduce the term 'prefix' that is used below)

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * The {@link #getPublisherBuilder(Config)} is called for every stream that needs to be created. The {@link Config} object
+ * passed to the method contains a subset of the global configuration, and with the prefixes removed. So for the previous
+ * configuration, it would be:
+ * <pre>
+ * type =  i.e.m.reactive.messaging.impl.kafka.Kafka
+ * bootstrap-services = localhost:9092
+ * topic = my-topic
+ * </pre>
+ * <p>
+ * So the connector implementation can retrieves the value with {@link Config#getValue(String, Class)} and
```
hutchig: retrieves -> retrieve

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * {@link Config#getOptionalValue(String, Class)}.
+ * <p>
+ * If the configuration is invalid, the {@link #getPublisherBuilder(Config)} method must throw an
+ * {@link IllegalArgumentException}, caught by the reactive messaging implementation and failing the deployment.
```
hutchig: "caught by the reactive messaging implementation and failing the deployment." ->
"caught by the reactive messaging implementation and triggering a deployment failure"
(do you want to say triggering a "DeploymentException"? - makes deployment more portable
across impls?

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * <p>
+ * Note that Reactive Messaging implementation must support the configuration format described here, implementation may
```
hutchig: "Note that Reactive Messaging implementations must support the configuration format described here. Implementations may support further, implementation specific, approaches."

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Incoming.java
cescoffier: I've replaced this with:

```
/**
     * The name of the consumed channel.
     *
     * @return the name of the consumed channel, must not be blank.
     */
```

which is more in line with the content of the spec.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
cescoffier: Done, in several places.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * deliver other approaches to create the streams.
+ */
+public interface IncomingConnectorFactory {
+
+    /**
+     * Gets the {@link MessagingProvider} class associated with this {@link IncomingConnectorFactory}.
+     * Note that the {@link MessagingProvider} is a user-facing interface used in the configuration.
+     *
+     * @return the {@link MessagingProvider} associated with this {@link IncomingConnectorFactory}. Must not be
+     * {@code null}. Returning {@code null} triggers a deployment failure.
```
hutchig: "triggers" -> "will cause"

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    Class<? extends MessagingProvider> type();
+
+    /**
+     * Creates a <em>stream</em> for the given configuration. The given configuration {@code type} attribute matches the
+     * {@link #type()}.
+     * <p>
+     * Note that the connection to the <em>transport</em> or <em>broker</em> is generally postponed until the
```
hutchig: "the subscription" -> "subscription occurs"

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+     * subscription.
+     *
+     * @param config the configuration, never {@code null}
+     * @return the created {@link PublisherBuilder}, must not be {@code null}.
```
hutchig: The 'never' and 'must not be' would be better the other way around as one is a constraint and the other is a promise. So parm 'must not be' and result 'will not be' would read best.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.reactive.messaging.Message;
+import org.eclipse.microprofile.reactive.messaging.MessagingProvider;
+import org.eclipse.microprofile.reactive.streams.operators.SubscriberBuilder;
+
+import java.util.NoSuchElementException;
+
+/**
+ * SPI used to implement a connector managing a sink of messages for a specific <em>transport</em>. Typically, to
```
hutchig: "Typically, " would be better as "For example, "

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * handle dispatch messages to Kafka, the reactive messaging extension would need to implement a {@code bean}
```
hutchig: I would remove "need to"

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * implementing this interface. This bean is called for every {@code stream} that need to be created for this specific
```
hutchig: "need" -> "needs"

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * <em>transport</em> (so Kafka in the example). These streams are connected to methods annotated with
```
hutchig: "so Kafka in the example" -> "(Kafka for example)"  or "(Kafka in this example)"

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * mp.messaging.outgoing.my-stream.type=i.e.m.reactive.messaging.impl.kafka.Kafka
+ * mp.messaging.outgoing.my-stream.bootstrap-servers=localhost:9092
+ * mp.messaging.outgoing.my-stream.topic=my-topic
+ * ...
+ * </pre>
+ * <p>
+ * This configuration keys are structured as follows: {@code mp.messaging.[incoming|outgoing].stream-name.attribute}.
+ * <p>
+ * The {@code stream-name} segment in the configuration key corresponds to the name of the stream used in the
+ * {@code Incoming} annotation:
```
hutchig: "Incoming" -> "Outgoing"

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * <pre>
+ * &#64;Outgoing("my-stream")
+ * public CompletionStage&lt;String&gt; produce(String s) {
+ *      // ...
+ * }
+ * </pre>
+ * <p>
+ * The set of attributes depend on the connector and transport layer (typically, bootstrap-servers is Kafka specific). The
```
hutchig: "typically" -> "for example"

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * {@code type} attribute is mandatory and indicates the fully qualified name of the {@link MessagingProvider}
+ * implementation. It must match the class returned by the {@link #type()} method. This is how a reactive messaging
+ * implementation looks for the specific {@link OutgoingConnectorFactory} required for a stream. In the previous
+ * configuration, the reactive extension implementation would need to find the {@link OutgoingConnectorFactory} returning
+ * the {@code i.e.m.reactive.messaging.impl.kafka.Kafka} class as result to its {@link #type()} method to create the
```
hutchig: "as result to" -> "as the result of"

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ *      // ...
+ * }
+ * </pre>
+ * <p>
+ * The set of attributes depend on the connector and transport layer (typically, bootstrap-servers is Kafka specific). The
+ * {@code type} attribute is mandatory and indicates the fully qualified name of the {@link MessagingProvider}
+ * implementation. It must match the class returned by the {@link #type()} method. This is how a reactive messaging
+ * implementation looks for the specific {@link IncomingConnectorFactory} required for a stream. In the previous
+ * configuration, the reactive extension implementation would need to find the {@link IncomingConnectorFactory} returning
+ * the {@code i.e.m.reactive.messaging.impl.kafka.Kafka} class as result to its {@link #type()} method to create the
```
hutchig: "result to" -> "the result of"

core-tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/connector/Dummy.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.reactive.messaging.tck.connector;
+
+import org.eclipse.microprofile.reactive.messaging.MessagingProvider;
+
+public class Dummy implements MessagingProvider {
```
hutchig: I think if we are going to keep it? we need to explain why it exists (to me too :-) )

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * using MicroProfile Config. The following snippet gives an example for a hypothetical Kafka connector:
+ *
+ * <pre>
+ * mp.messaging.outgoing.my-stream.type=i.e.m.reactive.messaging.impl.kafka.Kafka
+ * mp.messaging.outgoing.my-stream.bootstrap-servers=localhost:9092
+ * mp.messaging.outgoing.my-stream.topic=my-topic
+ * ...
+ * </pre>
+ * <p>
+ * This configuration keys are structured as follows: {@code mp.messaging.[incoming|outgoing].stream-name.attribute}.
```
hutchig: "Reactive messaging keys for OutgoingConnectionFactories are structured as follows: {@code mp.messaging.outgoing.stream-name.attribute}. "

In this context the [incoming/outgoing] will all be 'outgoing'? All??

If there is a read/write connector with some properties relavent to both
reading and writing are we asking user to duplicate these keys
once for incoming and once for outgoing? This seems less than
brilliant.

Also what is a user has LOTs of streams can we not have some
format that is for ALL streams of a certain type? To prevent
duplication?

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
hutchig: ...of course we are only using 'incoming' here.
Is this 'strictly' true, requiring the duplicate of some keys perhaps?

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * The set of attributes depend on the connector and transport layer (typically, bootstrap-servers is Kafka specific). The
+ * {@code type} attribute is mandatory and indicates the fully qualified name of the {@link MessagingProvider}
+ * implementation. It must match the class returned by the {@link #type()} method. This is how a reactive messaging
+ * implementation looks for the specific {@link OutgoingConnectorFactory} required for a stream. In the previous
+ * configuration, the reactive extension implementation would need to find the {@link OutgoingConnectorFactory} returning
+ * the {@code i.e.m.reactive.messaging.impl.kafka.Kafka} class as result to its {@link #type()} method to create the
+ * {@code my-stream} <em>subscriber</em>. Note that if the connector cannot be found, the deployment must be failed.
+ * <p>
+ * The {@link #getSubscriberBuilder(Config)} is called for every <em>sink</em> that needs to be created. The
+ * {@link Config} object passed to the method contains a subset of the global configuration, and with the prefixes removed.
```
hutchig: For the production of thus subset I would like to discuss a filter that can:
wildcarding for the stream-name
wildcard/either for incoming/outgoing

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * So for the previous configuration, it would be:
+ * <pre>
+ * type =  i.e.m.reactive.messaging.impl.kafka.Kafka
+ * bootstrap-services = localhost:9092
+ * topic = my-topic
+ * </pre>
+ * <p>
+ * So the connector implementation can retrieves the value with {@link Config#getValue(String, Class)} and
```
hutchig: 'retrieves' -> 'retrieve'

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * {@link Config#getOptionalValue(String, Class)}.
+ * <p>
+ * If the configuration is invalid, the {@link #getSubscriberBuilder(Config)} method must throw an
+ * {@link IllegalArgumentException}, caught by the reactive messaging implementation and failing the deployment.
```
hutchig: triggering a failure in the deployment.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * <p>
+ * Note that Reactive Messaging implementation must support the configuration format described here, implementation may
```
hutchig: "Note that a Reactive Messaging implementation must support the configuration format described here. Implementations are free to provide additional support for other approaches."

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+ * <p>
+ * The {@link #getPublisherBuilder(Config)} is called for every stream that needs to be created. The {@link Config} object
+ * passed to the method contains a subset of the global configuration, and with the prefixes removed. So for the previous
+ * configuration, it would be:
+ * <pre>
+ * type =  i.e.m.reactive.messaging.impl.kafka.Kafka
+ * bootstrap-services = localhost:9092
+ * topic = my-topic
+ * </pre>
+ * <p>
```
hutchig: In this example, if 'topic' was missing as a configuration property, the Kafka connector
would be at liberty to default to the stream name indicated in the annotation as the Kafka topic.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
cescoffier: You generally don't have the same channel name for incoming and outgoing. 

If you need something shared, you can use MP Config directly as the connector is a bean, it can retrieve the configuration. For instance, the Kafka connector could contain:

```
@Inject @ConfigProperty(name="bootstrap.servers") String servers; 
```

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
cescoffier: as the stream-name would likely be different, I don't see how you would do it. NEed to be discussed during the hangout.

cescoffier:
> 1 - wildcarding for properties between incoming/outgoing and across streams of one provider?

While it looks important, it's actually not that useful. Connectors being beans can consume the MP Config directly for common properties. The properties listed in the MP Reactive Messaging part are specific for a channel.

Note also that it's pretty rare to have the same channel name used for both incoming and outgoing.

> 2 - An @Outgoing annotation results in a Publisher but an OutgoingConnectionFactory
is used as a factory for Subscribers. I think this could get confusing for some users initially.

The name is actually for the end user and not for the Reactive Messaging implementer who, I hope, understands that to consume a stream you need to provide a `Subscriber`.

For the user, it matches `outgoing`. So if you are using ` `@Outgoing` in your code, you need to have a `mp.messaging.outgoing.channel` in your configuration.

> 3 - Number 2 above got me to thinking how much a 'OutgoingConnection' can be treated as a >Subscriber and a IncomingConnection as just another publisher. What is special about a Connection
> is that is has this means to be resolved into a synthetic Publisher/Subscriber by the reactive
> messaging specification. One could imagine a design where a interface
> MessagingProvider had methods
> getConnectorSubscriber
> getConnectorPublisher

`MessagingProvider` was just a marker interface. Note that some connectors implement only one side.   We can rename the methods from the `IncomingConnectorFactory` and `OutgoingConnectorFactory`, but they must be 2 interfaces. 

> This would seem better than having to load ALL the ConnectorFactories that
> can be loaded and then calling the type() method on (average 50%) of them
> to find a matching factory. It might be good to have lots of sources and syncs
> but have most of the classes not loaded if they are not used.

Good point. I never had this case because I expect the user to only include the one he wants to use (plus the dead code elimination techniques we use remove non-used connectors.)

We could image replacing the `type` method by a qualifier. 

> The 'connector' concept is really the notion of the system synthesizing
> in ingress/egress publisher/subscriber - it is strange that that concept
> is split into incoming and outgoing but that the MessagingProvider concept
> is not. There is no concrete reason why such a factory can only generate connections
> for one type of MessagingProvider - the logical thing is that a factory 'CAN'
> produce a connection for a messaging provider (as encoded in the 'type' of the
> annotation/config passed in. This enables a 'chain of responsibility' G04 pattern and allows
> for delegation etc. in the creation. Something for the hangout - so I will raise an issue - please
> don't let this issue prevent merge of the current PR.

If we remove the `MessagingProvider`, we would need a way to identify the connector. We can add a qualifier as indicated above. However, it must be a fully qualified name to avoid conflicts between simple names. 

Some connectors will only be able to implement one side. Typically: 

* mbus/dbus connector - only Incoming support
* Debezium - only outgoing support
* ML engine - only outgoing support

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
hutchig: i.e. set a property for all streams that use a particular provider. So for example if there is a property that is the same for all streams that use Kafka.
It seems to me that these properties will not be particularly easy to set up when there are lots of streams
and many of the streams will have identical properties that relate to the install or the users principle etc.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
hutchig: but would you not have to set this property with knowledge of what all the stream names were going to be? I don' think this scales well and may make it hard to seperate the concerns/config into application concerns (queue names) and devops concerns (rotating credentials etc)?

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
cescoffier: Then, they are in the connector bean configuration, not in the channel configuration. 
As the channel names are going to be different, what would be the prefix? 

`mp.messaging.[???]`

I don't really see the point as the connector can simply use:
```
@ConfigProperty(name="boostrap.servers") String servers;
```

And then the channels don't have to indicate the server anymore (only if they want another broker)

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
cescoffier: The queue names are definitely in the `mp.messaging....` properties. The rest depends on the connector implementation. BTW, it's very likely that each channel uses a different queue name.

hutchig: I think it would be highly useful to have properties that apply to a particular messaging provider. We can insert the queue name and incoming/ougoing into a synthetic property for Configs being passed to Connectors that match by the test of the key.

So for
mp.messaging.[incoming|outgoing].stream-name.attribute

If we know the stream-name is being processed by acme-kafka  MessagingProvider
then
mp.messaging.incoming.connector.acme-kafka.ticket = dsfsadf
can result in 
mp.messaging.incoming.stream-name.ticket = dsfsadf

Or we can just spec that a connector will take account of:

mp.messaging.[incoming|outgoing].[Connector-fully-qualified-classname].attribute

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,119 @@
...
+     * <p>
+     * Note that the connection to the <em>transport</em> or <em>broker</em> is generally postponed until the
+     * subscription occurs.
+     *
+     * @param config the configuration, must not be {@code null}
+     * @return the created {@link PublisherBuilder}, will not be {@code null}.
+     * @throws IllegalArgumentException if the configuration is invalid.
+     * @throws NoSuchElementException   if the configuration does not contain an expected attribute.
+     */
+    PublisherBuilder<? extends Message> getPublisherBuilder(Config config);
```
Emily-Jiang: Why should we pass in a config? The config object can be obtained by impl without any problem though.

Emily-Jiang: also can you move the spi to spi package?

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,109 @@
...
+     * <p>
+     * Note that the connection to the <em>transport</em> or <em>broker</em> is generally postponed until the
+     * subscription.
+     *
+     * @param config the configuration, never {@code null}
+     * @return the created {@link SubscriberBuilder}, must not be {@code null}.
+     * @throws IllegalArgumentException if the configuration is invalid.
+     * @throws NoSuchElementException if the configuration does not contain an expected attribute.
+     */
+    SubscriberBuilder<? extends Message, Void> getSubscriberBuilder(Config config);
```
Emily-Jiang: I don't think we should ask a config object to be passed over.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
hutchig: There is Servicability benefits if we trace the Config object we pass in and it contains what was used rather than encourage an approach that will make less likely Connector imples behave the same way? To take this further would be to allow every needed Connector property to be done in a custom way.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
cescoffier: This has been discussed during the hangout.

cescoffier:
I've switched the connector lookup using a `@Connector` qualifier (#30).
I also moved the connector interfaces to the spi package (as requested by @Emily-Jiang)

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/IncomingConnectorFactory.java
cescoffier: Done in my last push.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,112 @@
...
+ * handle the consumption of records from Kafka, the reactive messaging extension would need to implement a {@code bean}
+ * implementing this interface. This bean is called for every {@code channel} that needs to be created for this specific
+ * <em>transport</em> (so Kafka in this case). These channels are connected to methods annotated with
+ * {@link org.eclipse.microprofile.reactive.messaging.Incoming}.
+ * <p>
+ * The factory is called to create a {@code channel} for each configured <em>transport</em>. The configuration is done using
+ * MicroProfile Config. The following snippet gives an example for a hypothetical Kafka connector:
+ *
+ * <pre>
+ * mp.messaging.incoming.my-channel.type=i.e.m.reactive.messaging.impl.kafka.Kafka
```
hutchig: I think this attribute name should be "provider". This is because it allows us to comply
with the generic pattern for annotation parameter override of: 

mp.messaging.incoming|outgoing.channel-name.<parameter>=value

and the parameter in both annotations is called "provider". So we have
```
mp.messaging.incoming.my-channel.provider=value
```

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * <pre>
+ * &#64;Incoming("my-channel")
+ * public void consume(String s) {
+ *      // ...
+ * }
+ * </pre>
+ * <p>
+ * The set of attributes depend on the connector and transport layer (for example, bootstrap-servers is Kafka specific).
+ * The {@code type} attribute is mandatory and indicates the fully qualified name of the {@link MessagingProvider}
```
hutchig: The _type_ should be renamed to _provider_ and must be specified in at least one of: 

1. the annotation's _provider_ attribute
2. the relevant 'override' property name (e.g. mp.incoming.channel.provider)

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/Connector.java
```diff
@@ -0,0 +1,51 @@
...
+ * <p>
+ * The value indicates  the {@link MessagingProvider} class associated with  the bean implementing either
+ * {@link IncomingConnectorFactory} or {@link OutgoingConnectorFactory} or both.
+ * <p>
+ * Note that the {@link MessagingProvider} is a user-facing interface used in the configuration.
+ */
+@Qualifier
+@Retention(RUNTIME)
+@Target(TYPE)
+public @interface Connector {
```
hutchig: I am uncomfortable with a Connector( "value" ) where "value" is a Provider fqClassname and not a Connector fqClassName. 
I think we should call this annotation that matches the type name that its value is an instance of.

cescoffier:
> On 21 May 2019, at 17:22, Gordon Hutchison <notifications@github.com> wrote:
> 
> @hutchig commented on this pull request.
> 
> In api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/IncomingConnectorFactory.java:
> 
> > + * The portion of the key that precedes the {@code attribute} is acts as a property prefix that has a common structure
> + * across all {@code channels}.
> + * </p>
> + * <p>
> + * The {@code channel-name} segment in the configuration key corresponds to the name of the channel used in the
> + * {@code Incoming} annotation:
> + *
> + * <pre>
> + * &#64;Incoming("my-channel")
> + * public void consume(String s) {
> + *      // ...
> + * }
> + * </pre>
> + * <p>
> + * The set of attributes depend on the connector and transport layer (for example, bootstrap-servers is Kafka specific).
> + * The {@code type} attribute is mandatory and indicates the fully qualified name of the {@link MessagingProvider}
> The type should be renamed to provider and must be specified in at least one of:
> 
> the annotation's provider attribute
> the relevant 'override' property name (e.g. mp.incoming.channel.provider)

Agreed on the name. However for simplicity I would drop the provider attribute in the annotations.

cescoffier:
> On 21 May 2019, at 18:37, Gordon Hutchison <notifications@github.com> wrote:
> 
> @hutchig commented on this pull request.
> 
> In api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/Connector.java:
> 
> > +
> +import static java.lang.annotation.ElementType.TYPE;
> +import static java.lang.annotation.RetentionPolicy.RUNTIME;
> +
> +/**
> + * Qualifier used on connector implementations to indicate the associated {@link MessagingProvider}.
> + * <p>
> + * The value indicates  the {@link MessagingProvider} class associated with  the bean implementing either
> + * {@link IncomingConnectorFactory} or {@link OutgoingConnectorFactory} or both.
> + * <p>
> + * Note that the {@link MessagingProvider} is a user-facing interface used in the configuration.
> + */
> +@Qualifier
> +@Retention(RUNTIME)
> +@Target(TYPE)
> +public @interface Connector {
> I am uncomfortable with a Connector( "value" ) where "value" is a Provider fqClassname and not a Connector fqClassName.
> I think we should call this annotation that matches the type name that its value is an instance of.
> 
Using the connector FQN, means that the messaging provider becomes useless, and so should be removed. So we only have the connector class name.

cescoffier:
> On 21 May 2019, at 17:18, Gordon Hutchison <notifications@github.com> wrote:
> 
> @hutchig commented on this pull request.
> 
> In api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/IncomingConnectorFactory.java:
> 
> > +
> +import javax.enterprise.inject.spi.DeploymentException;
> +import java.util.NoSuchElementException;
> +
> +/**
> + * SPI used to implement a connector managing a source of messages for a specific <em>transport</em>. For example, to
> + * handle the consumption of records from Kafka, the reactive messaging extension would need to implement a {@code bean}
> + * implementing this interface. This bean is called for every {@code channel} that needs to be created for this specific
> + * <em>transport</em> (so Kafka in this case). These channels are connected to methods annotated with
> + * {@link org.eclipse.microprofile.reactive.messaging.Incoming}.
> + * <p>
> + * The factory is called to create a {@code channel} for each configured <em>transport</em>. The configuration is done using
> + * MicroProfile Config. The following snippet gives an example for a hypothetical Kafka connector:
> + *
> + * <pre>
> + * mp.messaging.incoming.my-channel.type=i.e.m.reactive.messaging.impl.kafka.Kafka
> I think this attribute name should be "provider". This is because it allows us to comply
> with the generic pattern for annotation parameter override of:
> 
> mp.messaging.incoming|outgoing.channel-name.=value
> 
> and the parameter in both annotations is called "provider". So we have
> 
> mp.messaging.incoming.my-channel.provider=value
Looking at the other comment, provider may become useless. So might be better to call is connector.

hutchig:
I agree that "messaging provider becomes useless," - it did use to provide a level of indirection
but this same can be acheived using, for example:
```
mp.messaging.incoming.channel-name.connector=[overriding-value]
```
If this seems OK I agree happily we can remove the Provider concept completely :-).
The only slight downside is that connectors interfaces are incoming/outgoing distinct whereas
providers were both. I guess having the same implementation class that implements both
interfaces means one can either have:
```
either
mp.messaging.connector.[fqClassname].attribute=value
or
mp.messaging.connector.[CdiQualifier].attribute=value

```
That will be applied to any outgoing OR incoming channel where `com.acme.Kafka`|`Kafka`
(fqn or qualifier) is the given `@Connector` value.

hutchig:
My questions are fully resolved/clarified now in #28 at https://github.com/eclipse/microprofile-reactive-messaging/issues/28#issuecomment-494666872

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/IncomingConnectorFactory.java
hutchig: This should now be:
```
mp.messaging.incoming.my-channel.connector=i.e.m.reactive.messaging.impl.kafka.Kafka
```

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,112 @@
...
+ * MicroProfile Config. The following snippet gives an example for a hypothetical Kafka connector:
+ *
+ * <pre>
+ * mp.messaging.incoming.my-channel.type=i.e.m.reactive.messaging.impl.kafka.Kafka
+ * mp.messaging.incoming.my-channel.bootstrap-servers=localhost:9092
+ * mp.messaging.incoming.my-channel.topic=my-topic
+ * ...
+ * </pre>
+ * <p>
+ * The configuration keys are structured as follows: {@code mp.messaging.[incoming|outgoing].channel-name.attribute}.
```
hutchig: The configuration keys are structured as follows: {@code mp.messaging.[incoming|outgoing].channel-name.attribute} or  {@code mp.messaging.connector.connector-qualifier-name.attribute}.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/Connector.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * @return the {@link MessagingProvider} associated with the bean implementing {@link IncomingConnectorFactory}
+     * or {@link OutgoingConnectorFactory}. Must not be {@code null}. Returning {@code null} will cause a deployment
+     * failure.
+     */
+    Class<? extends MessagingProvider> value();
```
hutchig: This will become `Class<? extends ConnectFactory>`  a superclass of incoming/outgoing?

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,112 @@
...
+ * <p>
+ * The {@code channel-name} segment in the configuration key corresponds to the name of the channel used in the
+ * {@code Incoming} annotation:
+ *
+ * <pre>
+ * &#64;Incoming("my-channel")
+ * public void consume(String s) {
+ *      // ...
+ * }
+ * </pre>
```
hutchig: <p>
The _connector-qualifier-name_ segment in the configuration key corresponds to the classname of the parameter passed to the {@code Connector} annotation that is used as a CDI qualifier in connector resolution. 
 *
 * <pre>
 * &#64;Connector(com.acme.Kafka.class)
 * &#64;Incoming("my-channel")
 * public void consume(String s) {
 *      // ...
 * }
 * </pre>

The above connector could be overridden with a property of, for example:
```
 mp.messaging.incoming.my-channel.connector=io.moresql.Rfewdbc
```
This would be equivalent of `@Connector( io.moresql.Rfewdbc.class)`

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,101 @@
...
+ * handle dispatch messages to Kafka, the reactive messaging extension would implement a {@code bean} implementing this
+ * interface. This bean is called for every {@code stream} that needs to be created for this specific <em>transport</em>
+ * (so Kafka in this example). These streams are connected to methods annotated with
+ * {@link org.eclipse.microprofile.reactive.messaging.Outgoing}.
+ * <p>
+ * The factory is called to create a {@code subscriber} for each configured <em>transport</em>. The configuration is done
+ * using MicroProfile Config. The following snippet gives an example for a hypothetical Kafka connector:
+ *
+ * <pre>
+ * mp.messaging.outgoing.my-channel.type=i.e.m.reactive.messaging.impl.kafka.Kafka
```
hutchig: 'type' -> 'connector'
and the value is the classname (without .class of course) of the default/value parameter of the Connector qualifier.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/Connector.java
hutchig: This is being resolved to be  Connector where the default/value parameter is
`Class<? extends ConnectorFactory>` where ConnectorFactory is a common
superclass of IncomingConnectorFactory and OutgoingConnectorFactory??

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/IncomingConnectorFactory.java
hutchig: If the `@Connector` annotation is not present and the _connector_ attribute is not set the channel is connected internally to the service. If either is present the channel is a connected channel and wired to an 
external message source or sink. If both are the `@Connector qualifier and the attribure are present, the attributes value will take precedence.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/OutgoingConnectorFactory.java
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * <pre>
+ * &#64;Outgoing("my-channel")
+ * public CompletionStage&lt;String&gt; produce(String s) {
+ *      // ...
+ * }
+ * </pre>
+ * <p>
+ * The set of attributes depend on the connector and transport layer (For example, bootstrap-servers is Kafka specific).
+ * The {@code type} attribute is mandatory and indicates the fully qualified name of the {@link MessagingProvider}
```
hutchig: If the @Connector annotation is not present and the connector attribute is not set the channel is connected internally to the service. If either is present the channel is a connected channel and wired to an
external message source or sink. If both are the `@connector qualifier and the attribure are present, the attributes value will take precedence. The value of this attribute is used to resolve a class that is used as the `@Connector` qualifier class' parameter value.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/IncomingConnectorFactory.java
```diff
@@ -0,0 +1,112 @@
...
+ * throwing a {@link DeploymentException} wrapping the exception.
+ * <p>
+ * Note that a Reactive Messaging implementation must support the configuration format described here. Implementations
+ * are free to provide additional support for other approaches.
+ */
+public interface IncomingConnectorFactory {
+
+    /**
+     * Creates a <em>channel</em> for the given configuration. The given configuration {@code type} attribute matches
+     * the {@link Connector} qualifier used on the bean.
```
hutchig: Creates a <em>channel</em> for the given configuration. The channel's configuration of the {@code connector} attribute can be used to override the {@link Connector} qualifier's parameter indicating the {@IncomingConnectorFactory} implementation class.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/Connector.java
cescoffier: To avoid classloading, I would just use:

```
String value();
```

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/IncomingConnectorFactory.java
cescoffier: My previous comment about using String instead of class would transform this to:

```
@Connector("acme.kafka")
```

and

```
mp.messaging.incoming.my-channel.connector=acme.kafka
```

Note that it would also impact connector _global_ config:

```
mp.messaging.acme.kafka.attr=value
```

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/IncomingConnectorFactory.java
cescoffier: Do you mean using `@Connector` with `@Incoming` or `@Outgoing`. While possible, I was not thinking about adding this in 1.0.

cescoffier:
> We need to resolve the Type extended by the Class used to define the type of the value parameter in the Connector qualifier. Perhaps a common superclass of the Incoming and Outgoing connection factories?

Or just use a String and problem gone :-D

hutchig:
OK - lets go with Connector( String ) for now to get this PR in and then if people want to refine further we can discuss.

cescoffier:
@hutchig I've addressed the comments, following the discussion from #22.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/connector/OutgoingConnectorFactory.java
hutchig: This will be handled in another PR

hutchig:
Feel free to merge from my POV.

PR #23, opened 2019-04-02
Add explicit reactive approach for message stream operation paragraph

### LINKED ISSUE #20: Direct call (through CDI interceptor) of @Outgoing methods
In https://gitter.im/eclipse/microprofile-reactive around the 4th of March
a user expected to be able to call a @Outgoing method from their own code.
 
Could we support this (to an Injected CDI proxy of course)
or alternatively how will we  better document that this is not an
expected behaviour that we support? 

Given this (mis)understanding he also had a method that took
a parameter (as his code could supply it) which might be
possible if user code were to call a (CDI proxy of a)  @Outgoing method.

hutchig:
This adds a small paragraph to clarify why it is not sensible
to call the annotated methods directly.

This PR resolves #20

### DIFF
--- a/spec/src/main/asciidoc/architecture.asciidoc
+++ b/spec/src/main/asciidoc/architecture.asciidoc
@@ -59,6 +59,18 @@ Note that the class is just a marker class, it provides no actual functionality.
 
 An application developer may specify a specific messaging provider using the `provider` property on the `@Incoming` and `@Outging` annotations.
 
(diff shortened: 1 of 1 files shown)


hutchig:
Is there anything controversial or needs rewording in this? Would anyone like to approve it?
(It is the same text I had merged in previously but this time via a more correct process.)

Issue #14, opened 2018-11-13
Clarify which CDI scopes are supported


mkouba:
For CDI beans that declare a subscriber/publisher/processor.

We should probably also clarify **"when"** the corresponding business methods are invoked.

Emily-Jiang:
`@ApplicationScoped`, `@SessionScoped`, `@ConversationScoped`, `@RequestScoped`, `@Dependent` beans need to be specified when the business methods are invoked.

cescoffier:
Depending on the signature, the method is called:
* the instance is created when the scope is initialized, the method is called immediately after that
* the instance is created when the scope is initialized, the method is called for every item transiting on the streams

This need to be clarified and enforced in the TCK.

mkouba:
`@SessionScoped`, `@ConversationScoped` and `@RequestScoped` do not make sense because an HTTP request needn't be active at the time the method is invoked.

hutchig:
+1 on @mkouba 's  @SessionScoped, @ConversationScoped and @RequestScoped do not make sense because an HTTP request needn't be active at the time the method is invoked."
As it stands....
    I THINK what people are suggesting is that IF there are enough outstanding request(N)
tokens then an incoming request could 'drive' a Publish from an @Outgoing - but if we we do this what will it mean when there is a request incoming and there are no tokens? and what means do we
use to connect up an incoming request to a reactive stream Publisher.
This is not the 'reactive' way and I think @mkouba 's questionw as answered by @cescoffier  without reference to the CDI @Request etc.  based scopes. 

Perhaps a subsequent spec level could tie in JVM/service ingress to reactive streams Publishers based on
feedback from Quarkus experience etc.

otaviojava:
IMHO: Just have one scope already defined such as `@ApplicationScoped`

cescoffier:
With #25 merged, the spec now contains a section about the supported scope.

PR #25, opened 2019-04-30
Add signatures, scope and acknowledgement sections to the spec

### LINKED ISSUE #1: Add scope (not CDI but other specs) section to spec.
The messaging spec should have a scope section that references the streams spec and its use of that.

cescoffier:
It also fixes #1, and #14.

### DIFF
--- a/spec/src/main/asciidoc/architecture.asciidoc
+++ b/spec/src/main/asciidoc/architecture.asciidoc
@@ -193,47 +193,441 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
 Enabling this would bypass the automatic back pressure mechanism that is one of the benefits of the specification.
 The `@Incoming` and `@Outgoing` method annotations are used to declaratively define the stream which is then run by the implementation of MicroProfile Reactive Messaging without the user's code needing to handle concerns such as subscriptions or flow control within the stream.
 
-==== Supported Bean scope
+==== Supported CDI scopes
 
+Implementations of the Reactive Messaging specification must support at least the following CDI scopes:
 
-==== Method signatures
+* `@ApplicationScoped` beans
+* `@Dependent` beans
 
-This section lists the method signatures that implementation must support.
+The following code gives an example of a bean annotated with `@ApplicationScoped`:
 
+[source,java]
+----
+@ApplicationScoped
+public class ApplicationScopeBeans {
+
+  @Outgoing("source")
+  public Publisher<Integer> source() {
+    return ReactiveStreams.of(id).buildRs();
+  }
+
+  @Incoming("source")
+  @Outgoing("output")
+  public int process(int i) {
+    return i + 1;
+  }
+
+  @Incoming("output")
+  public void sink(int v) {
+    System.out.println(v);
+  }
+}
+----
+
+Implementations can provide support for other scopes. 
+However the behavior is not defined.
+
+==== Supported method signatures
 
 The signature of message stream methods can have a number of different distinct types, offering differing levels of power and simplicity to application developers. Different shapes are supported depending on whether the method is a publisher, subscriber or processor, for example, a publishing stream supports returning MicroProfile Reactive Streams `PublisherBuilder`, but not `SubscriberBuilder`, the inverse is true for a subscribing stream.
 
+This section lists the methods signatures that must be supported by the Reactive Messaging implementation.
 Implementations must validate that the stream shape matches the `@Outgoing` and `@Incoming` annotations, if they don't, a CDI definition exception should be raised to the CDI container during initialization.
 
-===== Simple method streams
+It's important to remember that users must not call these methods directly. 
+They are invoked by the Reactive Messaging implementation following the Reactive Streams protocol.
+
+Also the method must be implemented as a non-blocking fashion.
+For blocking transformations, asynchronous variants can be used.
+
+INFO: _assembly_ time is when the Reactive Messaging implementation initializes itself and creates the different bean instances and connects them together.
+INFO: In the following lists, `Message` can be an implementation of the `Message` interface.
+
+===== Methods producing data
+
+[cols="2a,1,1",options="header"]
+|===
+|Signature
+|Behavior
+|Invocation
+
+|
+[source, java]
+----
+@Outgoing("name") 
+Publisher<Message<O>> method()
+----
+| Returns a stream of `Message` associated with the channel `name`. 
+| Method called once at assembly time.
+
+|
+[source, java]
+----
+@Outgoing("channel") 
+Publisher<O> method()
+----
+| Returns a stream of _payload_ associated with the channel `channel`. Produced payloads are mapped to `Message<O>` by the Reactive Messaging implementation.
+| Method called once at assembly time.
+
+|
+[source, java]
+----
+@Outgoing("channel") 
+PublisherBuilder<Message<O>> method()
+----
+| Returns a stream of `Message` associated with the channel `channel`. 
+| Method called once at assembly time.
+
+|
+[source, java]
+----
+@Outgoing("channel") 
+PublisherBuilder<O> method()
+----
+| Returns a stream of _payload_ associated with the channel `channel`. Produced payloads are mapped to `Message<O>` by the Reactive Messaging implementation.
+| Method called once at subscription time.
+
+|
+[source, java]
+----
+@Outgoing("channel") 
+Message<O> method()
+----
+| Produces a infinite streams of `Message` associated with the channel `channel`. 
+|	This method is called for each _request_ made by the subscriber.
+
+|
+[source, java]
+----
+@Outgoing("channel")
+O method()
+----
+| Produces an infinite streams of _payload_ associated with the channel `channel`. Produced payloads are mapped to `Message<O>` by the Reactive Messaging implementation.
+|	This method is called for each _request_ made by the subscriber.
+
+|
+[source, java]
+----
+@Outgoing("channel")
+CompletionStage<Message<O>> method()
+----
+| Produces a infinite streams of `Message` associated with the channel `channel`. The result is a `CompletionStage`. The method cannot be called until the `CompletionStage` returned previously is completed.
+|	This method is called for each _request_ made by the subscriber.
+
+|
+[source, java]
+----
+@Outgoing("channel") 
+CompletionStage<O> method()
+----
+| Produces a infinite streams of _payload_ associated with the channel `channel`. Produced payloads are mapped to `Message<O>` by the Reactive Messaging implementation. The result is a `CompletionStage`. The method cannot be called until the `CompletionStage` returned previously is completed.
+|	This method is called for each _request_ made by the subscriber.
+
+|===
+
+
+===== Methods consuming data
+
+[cols="2a,1,1",options="header"]
+|===
+|Signature
+|Behavior
+|Invocation
+
+|
+[source,java]
+----
+@Incoming("channel")
+Subscriber<Message<I>> method()
+----
+| Returns a `Subscriber` that receives the `Message` objects transiting on the channel `channel`.
+| The method is called only once to retrieve the `Subscriber` object at assembly time.
+ This subscriber is connected to the matching channel.
+
+|
+[source,java]
+----
+@Incoming("channel")
+Subscriber<I> method()
+----
+| Returns a `Subscriber` that receives the _payload_ objects transiting on the channel `channel`.
+The payload is automatically extracted from the inflight messages using `Message.getPayload()`.
+| The method is called only once to retrieve the `Subscriber` object at assembly time.
+ This subscriber is connected to the matching channel.
+
+|
+[source,java]
+----
+@Incoming("channel") 
+SubscriberBuilder<Message<I>> method()
+----
+| Returns a `SubscriberBuilder` that receives the `Message` objects transiting on the channel `channel`.
+| The method is called only once to retrieve the `Subscriber` object at assembly time. 
+The created `subscriber` is connected to the matching channel.
+
+|
+[source,java]
+----
+@Incoming("channel")
+SubscriberBuilder<I> method()
+----
+| Returns a `SubscriberBuilder` that receives the `Message` objects transiting on the channel `channel`.
+The payload is automatically extracted from the inflight messages using `Message.getPayload()`.
+| The method is called only once to retrieve the `Subscriber` object at assembly time.
+The created `subscriber` is connected to the matching channel.
+
+|
+[source,java]
+----
+@Incoming("channel")
+void method(I payload)
+----
+| Consumes the _payload_.
+The method can return `void` or any object or `null`. 
+The returned value is ignored. 
+| This method is called for every `Message<I>` instances transiting on the channel `channel`. 
+The payload is automatically extracted from the inflight messages using `Message.getPayload()`.
+The user method is never called concurrently and so must return before being called with the next payload.
+
+|
+[source,java]
+----
+@Incoming("channel")
+CompletionStage<?> method(Message<I> msg)
+----
+| Consumes the `Message` 
+| This method is called for every `Message<I>` instances transiting on the channel `channel`. 
+The user method is never called concurrently and so the returned `CompletionStage` must be completed before being calling the method with the next `Message`.
+Note that `@Incoming("channel") void method(Message<I> msg)` is not allowed as message acknowledgement is asynchronous.
+
+|
+[source,java]
+----
+@Incoming("channel") 
+CompletionStage<?> method(I payload)
+----
+| Consumes the _payload_ asynchronously
+| This method is called for every `Message<I>` instances transiting on the channel `channel`. 
+The payload is automatically extracted from the inflight messages using `Message.getPayload()`.
+The user method is never called concurrently and so the returned `CompletionStage` must be completed before being calling the method with the next `Message`.
+
+|===
+	
+===== Methods processing data
+
+[cols="2a,1,1",options="header"]
+|===
+|Signature
+|Behavior
+|Invocation
+
+| 
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out")
+Processor<Message<I>, Message<O>> method()
+----
+| Returns a Reactive Streams processor consuming incoming `Message` instances and produces `Message` instances.
+| This method is called once; at assembly time.
+
+| 
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+Processor<I, O> method();
+----
+| Returns a Reactive Streams processor consuming incoming _payload_ instances and produces _payload_ instances.
+| This method is called once; at assembly time.
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+ProcessorBuilder<Message<I>, Message<O>> method();
+----
+| Returns a `ProcessorBuilder` consuming incoming `Message` instances and produces `Message` instances.
+| This method is called once; at assembly time.
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+ProcessorBuilder<I, O> method();
+----
+| Returns a Reactive Streams processor that consuming incoming _payload_ instances and produces _payload_ instances.
+| This method is called once; at assembly time.
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+Publisher<Message<O>> method(Message<I> msg)
+----
+| Returns a Reactive Streams `Publisher` for each incoming `Message`.
+The returned `Publisher` can be empty or emits multiple `Message` instances.
+If the returned `Publisher` emits several elements, these elements are _flatten_ in the outgoing stream as a concatenation of elements.
+| This method is called for every incoming message. Never concurrently, so the stream returned previously must complete before having the method invoked with the next incoming `Message`.
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+Publisher<O> method(I payload)
+----
+| Returns a Reactive Streams `Publisher` for each incoming _payload_.
+The returned `Publisher` can be empty or emits multiple _payload_ instances.
+If the returned `Publisher` emits several elements, these elements are _flatten_ in the outgoing stream as a concatenation of elements.
+The Reactive Messaging implementation must create new `Message` instances for each emitted _payloads_ as well as extracing the payload for each incoming `Message`.
+| This method is called for every incoming message. Never concurrently, so the stream returned previously must complete before having the method invoked with the next incoming `Message`.
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+PublisherBuilder<Message<O>> method(Message<I> msg)
+----
+| Returns a `PublisherBuilder` for each incoming `Message`.
+The returned `PublisherBuilder` can be empty or emits multiple `Message` instances.
+If the returned `PublisherBuilder` emits several elements, these elements are _flatten_ in the outgoing stream as a concatenation of elements.
+| This method is called for every incoming message. Never concurrently, so the stream returned previously must complete before having the method invoked with the next incoming `Message`.
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+PublisherBuilder<O> method(I payload)
+----
+| Returns a `PublisherBuilder` for each incoming _payload_.
+The returned `PublisherBuilder` can be empty or emits multiple _payload_ instances.
+If the returned `PublisherBuilder` emits several elements, these elements are _flatten_ in the outgoing stream as a concatenation of elements.
+The Reactive Messaging implementation must create new `Message` instances for each emitted _payloads_ as well as extracing the payload for each incoming `Message`.
+| This method is called for every incoming message. Never concurrently, so the stream returned previously must complete before having the method invoked with the next incoming `Message`.
+
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+Message<O> method(Message<I> msg)
+----
+| Returns a `Message` for each incoming `Message`.
+| This method is called for every incoming message. Never concurrently, so the previous call must have returned before calling it with the next incoming `Message`.
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+O method(I payload)
+----
+| Returns a _payload_ for each incoming _payload.
+The Reactive Messaging implementation is responsible for unwrapping the _payload_ from the incoming `Message` and creating a `Message` from the returned _payload_.
+| This method is called for every incoming message. Never concurrently, so the previous call must have returned before calling it with the next incoming _payload_.
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+CompletionStage<Message<O>> 
+method(Message<I> msg)
+----
+| Produces a `Message` for each incoming `Message`. This method returns a `CompletionStage` that can redeem the `Message` instance asynchronously. The returned `CompletionStage` must not be completed with `null`.
+| This method is called for every incoming message. Never concurrently, so the previous call must have returned and produced a `Message` before calling it with the next incoming `Message`.
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+CompletionStage<O> method(I payload)
+----
+| Produces a _payload_ for each incoming _payload_. This method returns a `CompletionStage` that can redeem the _payload_ instance asynchronously. The returned `CompletionStage` must not be completed with `null`.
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,441 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
 The signature of message stream methods can have a number of different distinct types, offering differing levels of power and simplicity to application developers. Different shapes are supported depending on whether the method is a publisher, subscriber or processor, for example, a publishing stream supports returning MicroProfile Reactive Streams `PublisherBuilder`, but not `SubscriberBuilder`, the inverse is true for a subscribing stream.
 
+This section lists the methods signatures that must be supported by the Reactive Messaging implementation.
 Implementations must validate that the stream shape matches the `@Outgoing` and `@Incoming` annotations, if they don't, a CDI definition exception should be raised to the CDI container during initialization.
 
-===== Simple method streams
+It's important to remember that users must not call these methods directly. 
+They are invoked by the Reactive Messaging implementation following the Reactive Streams protocol.
+
+Also the method must be implemented as a non-blocking fashion.
```
hutchig: Also the method must be implemented in a non-blocking fashion.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,441 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+For blocking transformations, asynchronous variants can be used.
+
+INFO: _assembly_ time is when the Reactive Messaging implementation initializes itself and creates the different bean instances and connects them together.
+INFO: In the following lists, `Message` can be an implementation of the `Message` interface.
```
hutchig: This 'INFO' does not format to the start of a line (if that is the effect you wish).

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,441 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+| Returns a stream of `Message` associated with the channel `name`. 
+| Method called once at assembly time.
+
+|
+[source, java]
+----
+@Outgoing("channel") 
+Publisher<O> method()
+----
+| Returns a stream of _payload_ associated with the channel `channel`. Produced payloads are mapped to `Message<O>` by the Reactive Messaging implementation.
```
hutchig: regarding "Returns a stream of _payload_ associated with the channel `channel`." 
I understand that _payload_ is useda s the name of the parameter from "Message<T> of(T payload){ return ()-> payload; }" but in the text above this maps to "O". 
In the javadoc we use Message<T> so I think the above would be better as "Message<T>" (T for Type rather than O for Object). I think we should use the string 'payload' consistently - so that if it is used for the instance of the type T in the javadoc then we should only use it for the same thing here in the spec, for as an aid for those
trying to understand what is going on.

cescoffier: We cannot really use `<T>` as it would not work for `Processor` using 2 parameters `<I>` and `<O`>. 

I'm updating the sentence to say: `Returns a stream of _payload_ of type `O` associated with the channel `channel`.` Is it better?

spec/src/main/asciidoc/architecture.asciidoc
cescoffier: Missing empty line.... Fixed now.

spec/src/main/asciidoc/architecture.asciidoc
hutchig: Yes that is better. Thanks (and apologies for the delay in reply, I have off due to family issues).

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+| Method called once at subscription time.
+
+|
+[source, java]
+----
+@Outgoing("channel") 
+Message<O> method()
+----
+| Produces a infinite streams of `Message` associated with the channel `channel`.
```
hutchig: an infinite stream

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+| Produces a infinite streams of `Message` associated with the channel `channel`. 
+|	This method is called for each _request_ made by the subscriber.
+
+|
+[source, java]
+----
+@Outgoing("channel")
+O method()
+----
+| Produces an infinite streams of _payload_ associated with the channel `channel`. Produced payloads are mapped to `Message<O>` by the Reactive Messaging implementation.
```
hutchig: an infinate stream

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+|	This method is called for each _request_ made by the subscriber.
+
+|
+[source, java]
+----
+@Outgoing("channel")
+CompletionStage<Message<O>> method()
+----
+| Produces a infinite streams of `Message` associated with the channel `channel`. The result is a `CompletionStage`. The method cannot be called until the `CompletionStage` returned previously is completed.
```
hutchig: an infinite stream

hutchig: Is this specifying "the method should not be called by a reactive messaging implementation until..."?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+|	This method is called for each _request_ made by the subscriber.
+
+|
+[source, java]
+----
+@Outgoing("channel") 
+CompletionStage<O> method()
+----
+| Produces a infinite streams of _payload_ associated with the channel `channel`. Produced payloads are mapped to `Message<O>` by the Reactive Messaging implementation. The result is a `CompletionStage`. The method cannot be called until the `CompletionStage` returned previously is completed.
```
hutchig: an infinite stream

hutchig: Is this specifying "the method should not be called by a reactive messaging implementation until..."?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+ This subscriber is connected to the matching channel.
+
+|
+[source,java]
+----
+@Incoming("channel") 
+SubscriberBuilder<Message<I>> method()
+----
+| Returns a `SubscriberBuilder` that receives the `Message` objects transiting on the channel `channel`.
+| The method is called only once to retrieve the `Subscriber` object at assembly time.
```
hutchig: the `SubscriberBuilder` object

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+
+|
+[source,java]
+----
+@Incoming("channel") 
+SubscriberBuilder<Message<I>> method()
+----
+| Returns a `SubscriberBuilder` that receives the `Message` objects transiting on the channel `channel`.
+| The method is called only once to retrieve the `Subscriber` object at assembly time. 
+The created `subscriber` is connected to the matching channel.
```
hutchig: The `SubscriberBuilder` is used to build a `CompletionSubscriber` that is subscribed to the matching channel.

spec/src/main/asciidoc/architecture.asciidoc
hutchig: to retrieve the `CompletionSubscriber` object at assembly time.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+| The method is called only once to retrieve the `Subscriber` object at assembly time.
+ This subscriber is connected to the matching channel.
+
+|
+[source,java]
+----
+@Incoming("channel") 
+SubscriberBuilder<Message<I>> method()
+----
+| Returns a `SubscriberBuilder` that receives the `Message` objects transiting on the channel `channel`.
```
hutchig: that is used to build a `CompletionSubscriber` that

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+| The method is called only once to retrieve the `Subscriber` object at assembly time. 
+The created `subscriber` is connected to the matching channel.
+
+|
+[source,java]
+----
+@Incoming("channel")
+SubscriberBuilder<I> method()
+----
+| Returns a `SubscriberBuilder` that receives the `Message` objects transiting on the channel `channel`.
```
hutchig: that is used to build a CompletionSubscriber that

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+The payload is automatically extracted from the inflight messages using `Message.getPayload()`.
```
hutchig: The payload, of type I, is

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+| The method is called only once to retrieve the `Subscriber` object at assembly time.
```
hutchig: `SubscriberBuilder`

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+The created `subscriber` is connected to the matching channel.
```
hutchig: ..that is used to build a CompletionSubscriber that is subscribed to the matching channel.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+|
+[source,java]
+----
+@Incoming("channel")
+void method(I payload)
+----
+| Consumes the _payload_.
+The method can return `void` or any object or `null`. 
+The returned value is ignored. 
+| This method is called for every `Message<I>` instances transiting on the channel `channel`.
```
hutchig: every `Message<I>` instance

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+The user method is never called concurrently and so must return before being called with the next payload.
+
+|
+[source,java]
+----
+@Incoming("channel")
+CompletionStage<?> method(Message<I> msg)
+----
+| Consumes the `Message` 
+| This method is called for every `Message<I>` instances transiting on the channel `channel`.
```
hutchig: instance

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,442 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+
+|
+[source,java]
+----
+@Incoming("channel")
+CompletionStage<?> method(Message<I> msg)
+----
+| Consumes the `Message` 
+| This method is called for every `Message<I>` instances transiting on the channel `channel`. 
+The user method is never called concurrently and so the returned `CompletionStage` must be completed before being calling the method with the next `Message`.
```
hutchig: This is not actually a very logically clear statement IMO. The timeframe of the method invocation obviously can end prior to the CompletionStage being resolved. This is an `@Incoming` so we are providing the role of a Subscriber. We do not need the result from the CompletionStage in the same way that we would want the result from a Processor. Do we? Is there anyway that we want to be able to access the Streams CompletionSubscriber/CompletionStage in some middleware value add? If the user wishes to use a CompletionStage should we not allow multiple as yet unresolved Subscriber onData results?  Also, what about the result of the _completed_ stream - onComplete() is that just unavailable?

cescoffier: The `CompletionStage` is there to indicate when the incoming message has been processed correctly.  As the message can be asynchronously processed, the returned `CompletionStage` is used to know when the task has completed (or failed). 

But yes, the value used to complete the returned CompletionStage is meaningless (and discarded), and there is no way (at the moment) to retrieve the value. 

The completion of the `CompletionStage` controls the flow as we won't be calling this method before having received the completion of the previous call.

cescoffier: I've updated the content to explain more explicitly what's the expected behavior.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,438 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+Publisher<O> method(I payload)
+----
+| Returns a Reactive Streams `Publisher` for each incoming _payload_.
+The returned `Publisher` can be empty or emits multiple _payload_ instances.
+If the returned `Publisher` emits several elements, these elements are _flatten_ in the outgoing stream as a concatenation of elements.
```
hutchig: What function is used to do the flattening?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,438 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+PublisherBuilder<Message<O>> method(Message<I> msg)
+----
+| Returns a `PublisherBuilder` for each incoming `Message`.
+The returned `PublisherBuilder` can be empty or emits multiple `Message` instances.
+If the returned `PublisherBuilder` emits several elements, these elements are _flatten_ in the outgoing stream as a concatenation of elements.
```
hutchig: How do we flatten?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,438 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+| This method is called for every incoming message. Never concurrently, so the stream returned previously must complete before having the method invoked with the next incoming `Message`.
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+PublisherBuilder<O> method(I payload)
```
hutchig: Would this not be a ProcessorBuilder?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,438 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+----
+| Returns a `PublisherBuilder` for each incoming _payload_.
+The returned `PublisherBuilder` can be empty or emits multiple _payload_ instances.
```
hutchig: The stream resulting from the built Publisher can be...

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,438 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+If the returned `PublisherBuilder` emits several elements, these elements are _flatten_ in the outgoing stream as a concatenation of elements.
```
hutchig: If the stream emitted from the built Publisher...
Also the exact meaning of "_flatten_" is not clear to me - what is used to do the flattening?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,438 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+The Reactive Messaging implementation must create new `Message` instances for each emitted _payloads_ as well as extracing the payload for each incoming `Message`.
```
hutchig: _payload_ (singular) or 'each of the emitted _payloads_'

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,438 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+| This method is called for every incoming message. Never concurrently, so the stream returned previously must complete before having the method invoked with the next incoming `Message`.
```
hutchig: The use of the word "so" is awkward, I think we are saying. 
Implementations must not call the method subsequently until the stream emitted from the resolved Publisher is completed.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,438 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+Message<O> method(Message<I> msg)
+----
+| Returns a `Message` for each incoming `Message`.
+| This method is called for every incoming message. Never concurrently, so the previous call must have returned before calling it with the next incoming `Message`.
```
hutchig: As above.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,438 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
 
 The simplest shape that an application may use is a simple method. This is a method that accepts an incoming message, and possibly publishes an outgoing message:
 
 [source, java]
 ----
-@Incoming
-@Outgoing
+@Incoming("in")
+@Outgoing("out")
 public OutgoingMessage processMessage(IncomingMessage msg) {
```
hutchig: These have to be the same Message<type> don't they?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,438 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+| This method is called once, at _assembly_ time.
+
+|
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+PublisherBuilder<O> method(PublisherBuilder<I> pub)
+----
+| Transforms the incoming `PublisherBuilder` of _payload_. This method is used to manipulate streams and apply stream transformations.
```
hutchig: Apply a transformation to the stream represented by the `PublisherBuilder`.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,438 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
 [source, java]
 ----
-@Incoming
-@Outgoing
+@Incoming("in")
+@Outgoing("out")
 public OutgoingMessage processMessage(IncomingMessage msg) {
   return convert(msg);
 }
 ----
```
hutchig: Are the type names used IncomingMessage and OutgoingMessage slightly confusing when what is
actually being passed in is a _payload_? Suggest IncomingType and OutgoingType ?

hutchig: Or do we mean IncomingMessage implements Message<Type>?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -193,47 +193,438 @@ For example, calling an `@Outgoing` annotated method from user code will not pos
...
+If the method is not `@Outgoing` annotated, then the returned value is ignored - however, note that for asynchronous methods, the returned `CompletionStage` is still important for determining when message processing has completed successfully, for the purposes of message acknowledgement. 
+When there is no `@Outgoing` annotation, `void` may also be returned.
 
-===== Reactive streams
+===== Examples of methods using Reactive Streams or MicroProfile Reactive Streams Operators types
 
-For more power, developers may use Reactive Streams shapes. Reactive Streams shaped methods accept no parameters, and return one of the following:
+For more power, developers may use Reactive Streams instances. Reactive Streams shaped methods accept no parameters, and return one of the following:
 
 * `org.eclipse.microprofile.reactive.streams.PublisherBuilder`
```
hutchig: There are some of these in the section above.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -268,17 +650,349 @@ public PublisherBuilder<IncomingMessage, OutgoingMessage> processMessages() {
 
 ==== Message acknowledgement
 
+Acknowledgement is an important part of message processing.
 Messages are either acknowledged explicitly, or implicitly by the implementation.
 
-===== Explicit acknowledgement
+Acknowledgement is controlled by the `org.eclipse.microprofile.reactive.messaging.Acknowledgment` annotation.
```
hutchig: Acknowledgement for `@Incoming` messages is controlled by the

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -268,17 +650,349 @@ public PublisherBuilder<IncomingMessage, OutgoingMessage> processMessages() {
...
+
+* `NONE` - no acknowledgment is performed
+* `MANUAL` - the user is responsible for the acknowledgement, so the Reactive Messaging implementation does not apply implicit acknowledgement
+* `PRE_PROCESSING` - the Reactive Messaging implementation acknowledges the message before the annotated method or processing is executed
+* `POST_PROCESSING` - the Reactive Messaging implementation acknowledges the message once:
+
+  1. the method or processing completes if the method does not emit data
+  2. when the emitted data is acknowledged
+
+Each method signatures has different defaults and accept a different set of strategy.
```
hutchig: Each method signature type has different defaults and can implement different acknowledgement policies.

spec/src/main/asciidoc/architecture.asciidoc
hutchig: (I am not sure it is worth listing all the method shapes below that do not have an _@Outgoing_ annotation.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -268,17 +650,349 @@ public PublisherBuilder<IncomingMessage, OutgoingMessage> processMessages() {
...
+void method(I payload)
+----
+| Post-Processing	
+| None, Pre-Processing, Post-Processing
+
+|
+[source,java]
+----
+@Incoming("channel")
+CompletionStage<?> method(Message<I> msg)
```
hutchig: For shapes that return CompletionStage, for Post-Processing does the ack occur when the CompletionStage is completed or when the method returns?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -268,17 +650,349 @@ public PublisherBuilder<IncomingMessage, OutgoingMessage> processMessages() {
...
+| 
+[source,java]
+----
+@Incoming("in")
+@Outgoing("out") 
+Processor<I, O> method();
+----
+| Pre-Processing
+| None, Pre-Processing
+Post-Processing can be optionaly supported by implementations, however it requires a 1:1 mapping between the incoming message and the outgoing message.
```
hutchig: The RS spec refers to these as "elements signaled by the Subscriber/Publisher" I wonder if we should avoid using the word "message" for things that are not "Message<O>" type?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -268,17 +650,349 @@ public PublisherBuilder<IncomingMessage, OutgoingMessage> processMessages() {
...
 
+Acknowledgement is an important part of message processing.
 Messages are either acknowledged explicitly, or implicitly by the implementation.
 
-===== Explicit acknowledgement
+Acknowledgement is controlled by the `org.eclipse.microprofile.reactive.messaging.Acknowledgment` annotation. 
+The annotation allows configuring the acknowledgement strategy among:
+
+* `NONE` - no acknowledgment is performed
+* `MANUAL` - the user is responsible for the acknowledgement, so the Reactive Messaging implementation does not apply implicit acknowledgement
```
hutchig: the user is responsible for the acknowledgement, by calling the `Message#ack()` method,  so the..

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -268,17 +650,349 @@ public PublisherBuilder<IncomingMessage, OutgoingMessage> processMessages() {
...
+@Incoming("in")
+@Outgoing("out") 
+PublisherBuilder<O> method(PublisherBuilder<I> pub)
+----
+| Pre-Processing
+| None, Pre-Processing
+
+|===
+
+Invalid acknowledgement policies must be detected and a `DeploymentException` raised when the application starts.
```
hutchig: ...is deployed.        (It is possible to have a stopped but deployed application.)

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -317,14 +1032,9 @@ public Message<OutgoingMessage> processMessage(Message<IncomingMessage> msg) {
...
 ----
 
-===== Implicit acknowledgement
-
-Incoming messages may be implicitly acknowledged if the incoming message is not wrapped in `Message`. When it is acknowledged depends on the shape of the method. For simple method streams, the message should be acknowledged after the return value of the method is redeemed - for synchronous methods, this is when the method returns, for asynchronous methods, this is when the returned `CompletionStage` is redeemed.
-
-If the method returns a subscriber, then the message should be acknowledged before being sent to the subscriber.
-
-If the method returns a processor, then it is assumed that there is a 1:1 relationship between incoming and outgoing messages, and the implementation should acknowledge incoming messages as corresponding outgoing messages are published.
+The implementation is reponsible to enforce the acknowledgement strategy defined by the user when the `@Acknowledgement` policy is used.
```
hutchig: responsible for enforcing

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -317,14 +1032,9 @@ public Message<OutgoingMessage> processMessage(Message<IncomingMessage> msg) {
...
+The implementation is reponsible to enforce the acknowledgement strategy defined by the user when the `@Acknowledgement` policy is used. 
+If the annotation is not used, the default policy must be enforced.
 
-==== Custom message types
+// ==== Custom message types
 
-Implementations may provide custom subtypes of `Message` as an end user API, for the purpose of providing additional metadata associated with the messaging provider. If an application declares that it consumes or produces a subtype of the `Message` wrapper, the implementation should validate that the messaging provider supports that message wrapper type, and raise a deployment exception if it doesn't.
+// Implementations may provide custom subtypes of `Message` as an end user API, for the purpose of providing additional metadata associated with the messaging provider. If an application declares that it consumes or produces a subtype of the `Message` wrapper, the implementation should validate that the messaging provider supports that message wrapper type, and raise a deployment exception if it doesn't.
```
hutchig: Only minor comments. Apologies that it took me a number of days, I had an expected illness/death in the family and have been out of the office a lot recently.

spec/src/main/asciidoc/architecture.asciidoc
cescoffier: The used strategy is: `concatenation of the elements`. 
I'm going to add "The _flattening_ follows the same semantics as the `flatMap` operator from the MicroProfile Reactive Streams specification."

spec/src/main/asciidoc/architecture.asciidoc
cescoffier: I've rewritten it using `Message` and added a paragraph about implementations of `Message

spec/src/main/asciidoc/architecture.asciidoc
cescoffier: when the `CompletionStage` is completed.

cescoffier: actually, it's a bit more complicated, the ack happens when the emitted message is acknowledged, so definitely after the completion stage completion.

spec/src/main/asciidoc/architecture.asciidoc
cescoffier: It meant that `IncomingMessage` was implementing `Message`, but I've rewritten this section.

spec/src/main/asciidoc/architecture.asciidoc
cescoffier: I've removed the `@Outgoing` entries.

spec/src/main/asciidoc/architecture.asciidoc
cescoffier: I've switched to element.

cescoffier:
Thanks @hutchig!

PR #33, opened 2019-05-20
Add connector cross channel config properties

hutchig:

As discussed in today's hangout.


spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1051,21 +1051,29 @@ The Reactive Messaging implementation reads the global application configuration
 
 * `mp.messaging.incoming.[channel-name].[attribute]=[value]`
 * `mp.messaging.outgoing.[channel-name].[attribute]=[value]`
+* `mp.messaging.connector.[connector-name].[attribute]=[value]`
```
cescoffier: It's the associated messaging provider fully qualified name. I agree that connector-name is better here but it needs to be indicated somewhere.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1051,21 +1051,29 @@ The Reactive Messaging implementation reads the global application configuration
...
 
 For each extracted `channel-name`:
 
 1. The `type` attribute value is read, and the connector implementation identified. If no connector matches, the deployment must be failed with a `DeploymentException`;
-2. The other attributes are processed to generate a `Config` object containing only `attribute=value` entries.
+2. Relevant attributes are those matching either the `channel-name` or the resolved `connector-name`.
+3. Relevant attributes are processed to generate a `Config` object containing only `attribute=value` entries.
+If is valid to have an attribute specified at a connector level and also for a specific channel.
+If an attribute appears for both a channel and its relevant connector, the connector specific value will be used.
```
cescoffier: I would have said otherwise. If the property is defined at the connector level and at the channel level, the channel level wins no? 

Typically, you have a bunch of kafka channels, all but one use the same bootstrap servers, but one of them used another one. That would not be possible with connector > channel.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1051,21 +1051,29 @@ The Reactive Messaging implementation reads the global application configuration
...
 
 For example, the following snippet gives an example for a hypothetical Kafka connector:
  
 [source]
 ----
  mp.messaging.incoming.my-channel.type=org.eclipse.reactive.sample.kafka.Kafka
- mp.messaging.incoming.my-channel.bootstrap-servers=localhost:9092
  mp.messaging.incoming.my-channel.topic=my-topic
+ mp.messaging.connector.org.eclipse.reactive.sample.kafka.Kafka.bootstrap-servers=localhost:9092
```
cescoffier: bootstrap.servers (and not -)

spec/src/main/asciidoc/architecture.asciidoc
hutchig: Agh this is just a 'typo' more of a 'braino' as I agree with you!

spec/src/main/asciidoc/architecture.asciidoc
hutchig: I was thinking that the string used for the [connector-name] should match the value give for the `type` attribute. I agree that 'connector' is a much cleaner string to use in the literal [incoming|outgoing|connector] but I am less comfortable with that literal string being used to indicate the fqClassname of a MessagingProvider.  I think the root of the cause is that have both "Connector" and "MessagingProvider" as synonyms that mean the same thing but you have to 'learn' which is which is not a great design. 
We have:
1 - a marker interface for a Class used as the 'type' which is a key to a set of attributes.
2 - an injectable IncomingConnectorFactory
3 - an injectable OutgoingConnectorFactory
If we are using CDI qualifiers etc to resolve the connector, I wonder about using the
connecter bean to obtain the 'type'. This can be done in a more typesafe manner and
could mean we get rid of the MessagingProvider as a seperate entity and are left with
the simpler notion of a single concept - a connector class.

spec/src/main/asciidoc/architecture.asciidoc
hutchig: will fix

spec/src/main/asciidoc/architecture.asciidoc
hutchig: Done.

spec/src/main/asciidoc/architecture.asciidoc
hutchig: Done.

spec/src/main/asciidoc/architecture.asciidoc
hutchig: I note that we have got
```
Class<? extends MessagingProvider> provider() default MessagingProvider.class
```
as an attribute in both `@Outgoing` and '@Incoming' annotationsso we can have code like

@Outgoing( "orders_channel", provider="com.acme.Kafka")

I think it is a wise design choice if we allow all parameters of the `@Incoming/@Outgoing` annotation to be overridden with properties that begin: 
```
mp.messaging.incoming.[channel-name].[parameter-name] = [override to annotation value]
```
With no 'special' cases for type/provider 
In which case. 
As we currently have the 'provider' attribute of both the `@Incoming` and `@Outgoing`
annotations we need to allow propeties of: 
```
mp.messaging.outgoing.[channel-name].provider=[qualifier-value-or-fq-classname]
mp.messaging.incoming.[channel-name].provider=[qualifier-value-or-fq-classname]
```
If we add more parameters to the annotations these will be overridable by the same pattern.
Having a different design is less 'orthoganol' and requires user to learn 'special cases.'

hutchig: I would like to rename the `type()` method in `IncomingConnectionFactory` and `OutgoingConnectionFactory` to `provider()` or `getProvider()` I will see if I can
comment on that PR...

hutchig: Basically I have to see how #22 nets out and then come back and edit these comments on top of our use of Qualifiers or however else #22 nets out (as it will remove the provider annotation parameter!

hutchig:
HOLD label as don't merge until merging on top of #22 as that affect the property name provider|connector etc.

cescoffier:
@hutchig I've made another PR on the TCK (not related to this change).

hutchig:
I cannot yet write/test tck changes as I have not built our connector support!

hutchig:
Fixes 28 - well at least the concern that is left about having repeated similar properties for every channel of a connector.

cescoffier:
I will extend the TCK in another PR. 
Wondering if we need to update the `Factory` classes with this content also (it mentions the configuration format). WDYT?

cescoffier:
Merging, the rest of the work can be done in other PRs.

hutchig:
Thanks for merging - feel free to do this (from my POV) for any of my PRs that I request a review for that I have not put a HOLD label on.

hutchig:
I did the javadoc changes for the Factories under #42

PR #40, opened 2019-06-04
Add test to the TCK checking the new connector config format.

cescoffier:
As discussed during the hangout, here are a few tests around the new format.

### DIFF
--- a/core-tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/connector/FakeConfig.java
+++ b/core-tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/connector/FakeConfig.java
@@ -43,7 +43,7 @@ import java.util.*;
  *      mp.messaging.incoming.dummy-source-2.attribute=value-2
  *
  *      ## Dummy Outgoing connector
- *      mp.messaging.outgoing.dummy-sink.connector=oDummy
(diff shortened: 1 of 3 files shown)


core-tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/connector/FakeConfig.java
```diff
@@ -29,11 +29,19 @@
...
+ *
  *      ## Dummy Incoming connector
  *      mp.messaging.incoming.dummy-source.connector=Dummy
  *      mp.messaging.incoming.dummy-source.attribute=value
  *      mp.messaging.incoming.dummy-source.items=a,b,c,d,e,f,g,h,i,j
  *
+ *      ## Another Dummy Incoming channel
+ *      mp.messaging.incoming.dummy-source-2.connector=Dummy
+ *      mp.messaging.incoming.dummy-source-2.attribute=value-2
+ *
```
hutchig: Lines 46 and 75 don't mastch ion tyhat one is 'oDummy' and one is 'Dummy'.

cescoffier: nice catch!

PR #42, opened 2019-06-04
Add in connector attributes to the ConnectorFactory javadoc

hutchig:


hutchig:
One thing to note was that I removed a bit of text that said the connector attribute was manadatory. I think that we may wish to enable channels that are not 'connected' to have configuration as well
and these do not have associated connectors.

PR #44, opened 2019-06-06
Clarify the acknowledgement mechanism

cescoffier:
Fix #29.

This PR is the result of our discussions around acknowledgment:

* Forbid the usage of `POST_PROCESSING` when dealing with `Messages` - the user is responsible for the acknowledgment in this case.
* Clarify when the acknowledgment happens and how the chain of ack works
* Add more examples
* Fix a few typos

PR #45, opened 2019-06-07
Forbid post-processing ack where a Processor produces a Message

Azquelt:
If the user's processor produces the message, the reactive messaging
implementation cannot determine when that message is acknowledged
without wrapping it and hiding the actual class of the Message.

I think this was missing from #44


cescoffier:
good catch! That was a left-over.

PR #51, opened 2019-06-18
Remove use of raw Message type

### LINKED ISSUE #50: Raw Message type used in connector API
`OutgoingConnectorFactory` and `IncomingConnectorFactory` both have generic type parameters of `? extends Message`, resulting in a raw type warning.

I guess we should be using `? extends Message<?>` here since we really aren't specifying anything about the message payload?

Azquelt:
Fixes #50


Azquelt:
I _think_ this is all that's needed, but if I've missed something or there's some good reason for using raw types here, please reject.

PR #56, opened 2019-06-28
Fix TCK issues to do with application packaging

Azquelt:
This PR fixes the following TCK issues which affect the packaging of the test applications. Most of them only cause failures when running in a remote or managed arquillian container.

1) Package up the awaitility, assertJ and RXJava dependencies in the test app
2) Package the TckBase and ArchiveExtender classes in the test apps that depend on them
3) Remove FakeConfig and instead deploy the config in a `META-INF/microprofile-config.properties` file inside the test application. This is portable across MP Config implementations.
4) Switch from `@Test(expected = MyException)` to `@ShouldThrowException(MyException)`
    - `@ShouldThrowException` is an arquillian annotation which checks through the cause chain of a `DeploymentException` to see if the deployment failure was ultimately caused by the expected exception.


tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/ProcessorChainTest.java
```diff
@@ -39,7 +39,7 @@
     @Deployment
     public static Archive<JavaArchive> deployment() {
         JavaArchive archive = ShrinkWrap.create(JavaArchive.class)
-            .addClasses(BeanWithChain.class)
+            .addClasses(BeanWithChain.class, ArchiveExtender.class)
```
kenfinnigan: Not sure I understand why `ArchiveExtender` needs to be inside the deployment when it's used when creating the deployment, which happens outside the container

Azquelt: Yeah, I was a bit confused about that too.

The class can be loaded ok without `ArchiveExtender` being present but during application startup, open liberty was calling `Class.getDeclaredMethods()` on the test class as part of its injection processing. This reflection call threw a `NoClassDefFoundError` for `ArchiveExtender`.

Emily-Jiang: Moving to the deployment will ensure the tck run both by standalone TCK runner (different jvm from the runtime jvm) and the embedded TCK runner (same jvm as the one running the runtime). This issue only occurs when running in the standalone TCK runner mode.

PR #58, opened 2019-07-01
Add an unwrap method to the Message interface.

cescoffier:
This would allow users of a connector to avoid casting. Instead, the user would use the `unwrap` method. For instance: `KafkaMessage myMessage = msg.unwrap(KafkaMessage.class)`.

The default implementation only supports `Message` as parameter. Connector implementation would need to override the method.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
+++ b/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
@@ -80,4 +80,29 @@ public interface Message<T> {
     default CompletionStage<Void> ack() {
         return CompletableFuture.completedFuture(null);
     }
+
+    /**
+     * Returns an object of the specified type to allow access to the connector-specific {@link Message} implementation.
+     * If the {@link Message} implementation does not support the target class, an {@link IllegalArgumentException}
+     * should be raised.
+     * The default implementation only supports the {@link Message} class as argument. When a connector provides
+     * its own {@link Message} implementation, it should override this method to support the specific type.
+     *
+     * @param unwrapType the class of the object to be returned, must not be {@code null}
+     * @param <C> the target type
+     * @return an instance of the specified class
+     * @throws IllegalArgumentException if the current {@link Message} instance does not support the call
+     */
+    @SuppressWarnings({"unchecked"})
+    default <C> C unwrap(Class<C> unwrapType) {
+        if (unwrapType == null) {
+            throw new NullPointerException("The target class must not be `null`");
+        }
+        if (Message.class.equals(unwrapType)) {
+            return (C) this;
+        }
+        throw new IllegalArgumentException("Cannot unwrap an instance of " + this.getClass().getName()
+            + " to " + unwrapType.getName());
+
+    }
 }
--- a/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
+++ b/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
@@ -96,13 +96,16 @@ public interface Message<T> {
     @SuppressWarnings({"unchecked"})
     default <C> C unwrap(Class<C> unwrapType) {
         if (unwrapType == null) {
-            throw new NullPointerException("The target class must not be `null`");
+            throw new IllegalArgumentException("The target class must not be `null`");
         }
-        if (Message.class.equals(unwrapType)) {
-            return (C) this;
+        try {
+            return unwrapType.cast(this);
         }
-        throw new IllegalArgumentException("Cannot unwrap an instance of " + this.getClass().getName()
-            + " to " + unwrapType.getName());
+        catch (ClassCastException e) {
+            throw new IllegalArgumentException("Cannot unwrap an instance of " + this.getClass().getName()
+                + " to " + unwrapType.getName(), e);
+        }
+
 
     }
 }
--- a/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
+++ b/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
@@ -82,11 +82,14 @@ public interface Message<T> {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
```diff
@@ -80,4 +80,29 @@ public T getPayload() {
...
+     * @param <C> the target type
+     * @return an instance of the specified class
+     * @throws IllegalArgumentException if the current {@link Message} instance does not support the call
+     */
+    @SuppressWarnings({"unchecked"})
+    default <C> C unwrap(Class<C> unwrapType) {
+        if (unwrapType == null) {
+            throw new NullPointerException("The target class must not be `null`");
+        }
+        if (Message.class.equals(unwrapType)) {
```
Azquelt: Should this be an `instanceof` check?

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
```diff
@@ -80,4 +80,29 @@ public T getPayload() {
...
+            return (C) this;
+        }
+        throw new IllegalArgumentException("Cannot unwrap an instance of " + this.getClass().getName()
+            + " to " + unwrapType.getName());
```
Azquelt: Could also consider just calling `Class.cast()` catching the cast exception and creating your `IllegalArgumentException`?

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
cescoffier: and `instanceOf` would always return `true` for any Message implementation.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
cescoffier: implementations may not use a cast, but here we can.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
Azquelt: Doh, I totally misread that...

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
cescoffier: If we use:

```
try {
            return unwrapType.cast(this);
        }
        catch (ClassCastException e) {
            throw new IllegalArgumentException("Cannot unwrap an instance of " + this.getClass().getName()
                + " to " + unwrapType.getName(), e);
        }
```

No more issue with the `.equals`

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
```diff
@@ -80,4 +80,29 @@ public T getPayload() {
...
+     *
+     * @param unwrapType the class of the object to be returned, must not be {@code null}
+     * @param <C> the target type
+     * @return an instance of the specified class
+     * @throws IllegalArgumentException if the current {@link Message} instance does not support the call
+     */
+    @SuppressWarnings({"unchecked"})
+    default <C> C unwrap(Class<C> unwrapType) {
+        if (unwrapType == null) {
+            throw new NullPointerException("The target class must not be `null`");
```
Emily-Jiang: I suggest to throw IllegalArgumentException instead of NPE.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
Azquelt: Ah yes, ok, I read the above line `return (C) this` and thought we were casting ourselves to `C`, but in fact you're only doing that if `C` is actually `Message`

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
cescoffier: Done!

cescoffier:
@Azquelt I've made the suggested change.

Azquelt:
Ok, the code looks fine. Just a couple more questions.

Why "unwrap"?

I was thinking about a `Message` as an object which wraps a payload, so "unwrap" suggests that I would be getting the payload out of the message. However, what we're actually doing is getting a more specific message.

I am happy to go with `unwrap` if you think that's best.

Also, `C` isn't limited to implementations of `Message`. For example, I could potentially implement a `Message` that could be unwrapped to a kafka `ConsumerRecord`. Would that be acceptable behaviour or not?

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
Azquelt: I was so slow understanding the code that by the time I finished writing it, you'd changed it to how I thought it worked in the first place.

If you change it like this, does the javadoc need updating to say that the default implementation will try to cast the object to the requested type?

cescoffier:
I picked the `unwrap` name because it's used in other specs. For example: 

* https://github.com/javaee/jpa-spec/blob/master/javax.persistence-api/src/main/java/javax/persistence/EntityManagerFactory.java#L211.
* https://javaee.github.io/javaee-spec/javadocs/index.html?javax/validation/valueextraction/class-use/Unwrapping.html
* it's been discussed in MP Config (https://github.com/eclipse/microprofile-config/issues/84)

And about your second question, yes it's possible to unwrap to a non-message implementation (connector specific), such as you mentioned, a `ConsumerRecord` or a `ProducerRecord` for Kafka.

Azquelt:
Thanks, that makes it make more sense to me.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
cescoffier: I've updated the javadoc.

PR #61, opened 2019-07-12
made @Connector applicable to fields, methods, etc, since it's a Qualifier

michalszynkiewicz:


cescoffier:
What would be the use case? 

`@Connector` is a qualifier indicating the name of the connector implementation. By doing this, you would be able to inject the `Connector` bean in another bean (which is an interesting use case BTW)

michalszynkiewicz:
The ability to inject a connector is exactly what I'm targeting here.

PR #66, opened 2019-09-02
Fix #63 - Clarify the NONE ack policy usage

### LINKED ISSUE #63: How do we tell Connectors about config state such as `@Acknowledgement(None)`
It would be useful to tell a Connector what the annotations for a method are.

Some specs have a way of overriding Annotations using Config properties and there
is a bit of precedance in the mapping.
e.g.
https://github.com/eclipse/microprofile/pull/100/files
`<fully qualified class name>/<field name>/<spec-defined annotation name>/<annotation attribute name>`.

Could we run this in reverse to arrive at a property that descibes the annotations
that could be passed to the connector.

Just an idea.

cescoffier:


spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -699,6 +698,10 @@ public Message<Integer> process(KafkaMessage<String, Integer> input) {
 }
 ----
 
+The `NONE` acknowledgment strategy indicates that the message is not acknowledged by the method (neither implicitly nor explicitly).
+However, note that all incoming messages must be acknowledged, so using `NONE` indicates that the message is acknowledged in a different location or using a different mechanism.
```
hutchig: This could benefit from something more to differentiate it from "MANUAL:managed by the user code". Do you mean that myMessage.ack() still needs to be called but just not in this method or is it referring to Kafka layer acknowledgement must still be done for example?

hutchig: It is difficult to explain - without introducing the concept that we may be sending 1 message to more than 1 receiver (which is a concept that is not in the spec currently).

hutchig: If the difference is in effect then it might help to explain what the effect ( of having NONE versus MANUAL) is expected to be. For example.
"Specifying 'NONE' allows for this method to complete processing and return without handling acknowledgement and for this to be considered valid behaviour of the method. 
However, if messages were never acknowledged this would result in
a build up of unacknowledged messages in the system as no
automatic acknowledgement will be done if if 'NONE' is specified.
If 'NONE' is specified, each of the message object's ack() method should still be invoked
once as part of the overall processing, either before, during or after this method's execution."

cescoffier: > "Specifying 'NONE' allows for this method to complete processing and return without handling acknowledgment and for this to be considered valid behavior of the method.

Yes. 

> When 'NONE' is specified, each message object's ack() method should still be invoked
once as part of the overall processing but this is considered valid behaviour
either before, during or after this method's execution."

Exactly

I'm going to update the PR with this content.

cescoffier:
@hutchig I've updated the PR with your text.

Issue #76, opened 2019-11-06
InvalidConfigurationTest#incompleteChain tests BeanWithBadOutgoingSignature


danielkec:
instead of BeanWithIncompleteChain

cescoffier:
CC @lamtrhieu

lamtrhieu:
Thanks @cescoffier Let me have a look and try to fix this

cescoffier:
Thanks! Let me know if you need anything.

PR #77, opened 2019-11-08
Fix incorrect bean name, issue #76

lamtrhieu:
Fix https://github.com/eclipse/microprofile-reactive-messaging/issues/76

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/invalid/InvalidConfigurationTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/invalid/InvalidConfigurationTest.java
@@ -74,7 +74,7 @@ public class InvalidConfigurationTest {
   @ShouldThrowException(value = DeploymentException.class, testable = true)
   public static Archive<JavaArchive> incompleteChain() {
     JavaArchive archive = ShrinkWrap.create(JavaArchive.class)
-      .addClasses(BeanWithBadOutgoingSignature.class, ArchiveExtender.class)
+      .addClasses(BeanWithIncompleteChain.class, ArchiveExtender.class)
       .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml");
 
     ServiceLoader.load(ArchiveExtender.class).iterator().forEachRemaining(ext -> ext.extend(archive));


lamtrhieu:
@cescoffier Can you please have a quick look if the fix is ok. One problem is that I cannot run the test on my local computer. 
My intellij complain "Arquillian test class should have only 1 method with Deployment annotation" and when I run the test I cannot see any result. 
Can you please advise how can I fix that and run the tests?
Thanks

cescoffier:
The fix looks good to me.

You need to sign the Eclipse CLA and sign-off your commit.

lamtrhieu:
Yes, I know that. But can you show me how I can run the tests on my local computer ?
There seems like error in my IntelliJ with the error "Arquillian test class should have only 1 method with Deployment annotation"

cescoffier:
@lamtrhieu you need to test again an implementation. Build the package and update the smallrye implementation to use your tck artifact.

lamtrhieu:
@cescoffier I have successfully built the package now. How can I test smallrye implementation using the new tck package? I have checked out project **smallrye-reactive-messaging** already. 
I guess I need to edit the pom file and run some mvn command ? 
Can you please help me with the steps ? Thank you so much for that.

cescoffier:
You just need to change the MicroProfile Reactive Messaging version in the `pom.xml` file and run `mvn clean install`. 

The fast track would be to just build the `tck` module, that should be enough.

lamtrhieu:
Hi @cescoffier  
I build the tck project using my updated tck artifact and it get error with that test

"java.lang.RuntimeException: Expected exception of type javax.enterprise.inject.spi.DeploymentException during deployment of incomplete-chain, but no exception was thrown."

It's clear that no exception is thrown when deploy the bean. Can you please help me to understand why the bean is named "BeanWithIncompleteChain" and why it should throw exception with that kind of bean ?

cescoffier:
so... you just found a feature :-)

I was able to get the TCK passing by enabling the strict mode of smallrye reactive messaging (System.setProperty(STRICT_MODE_PROPERTY, "true");).

It means that we are all good with your PR.

cescoffier:
Thanks!

lamtrhieu:
@cescoffier If there any problem that need a little more "lines" to fix then just tag me in. I am happy to have a look.

PR #80, opened 2019-11-14
Clarify @Incoming method signatures

### LINKED ISSUE #79: Tck SubscriberShapeTest outside of spec
Signature of method [SubscriberBean#consumePayloadsAndReturnSomething](https://github.com/eclipse/microprofile-reactive-messaging/blob/a89bdc57f917827deba171bbea1b067a4aa8c171/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/signatures/subscribers/SubscriberBean.java#L129) is outside of spec defined variants.
```java
    @Incoming("string-payload")
    public String consumePayloadsAndReturnSomething(String payload) {
        increment("string-payload");
        add("string-payload", payload);
        return payload;
    }

```
Spec states:
`Implementations must validate that the stream shape matches the @Outgoing and
@Incoming annotations, if they don’t, a CDI definition exception should be raised to the CDI container
during initialization.`

cescoffier:
Fix #79

* Remove tests using non-supported signatures
* Add new tests in the TCK to be sure these non-supported signatures are detected and reported using CDI Definition Exception
* Clarify the spec about these signatures

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/signatures/invalid/InvalidSubscriberSignatureTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/signatures/invalid/InvalidSubscriberSignatureTest.java
@@ -1,0 +1,75 @@
+/**
+ * Copyright (c) 2018-2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/signatures/invalid/InvalidSubscriberSignatureTest.java
```diff
@@ -0,0 +1,75 @@
...
+        JavaArchive archive = ShrinkWrap.create(JavaArchive.class)
+            .addClasses(IncomingReturningNonVoidCompletionStage.class, ArchiveExtender.class)
+            .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml");
+
+        ServiceLoader.load(ArchiveExtender.class).iterator().forEachRemaining(ext -> ext.extend(archive));
+        return archive;
+    }
+
+    @Test
+    public void checkThatIncomingShouldNotReturnNonVoidCompletionStagw() {
```
xstefank:
```suggestion
    public void checkThatIncomingShouldNotReturnNonVoidCompletionStage() {
```

cescoffier: good catch! 
Updated.

PR #92, opened 2020-01-31
Fixes to the MetricsTestBean

### LINKED ISSUE #91: MetricsTestBean issues
The `MetricsTestBean` class has a few issues:

* There is a `System.out` statement in the `simpleMapping` test
* `inAppMessagesReceived` should be an `AtomicInteger` instead of `volatile`. The `++` operation is not atomic.

Azquelt:
* Change volatile int to an AtomicInteger
* Remove debug println

Fixes #91


cescoffier:
Thanks @Azquelt !

PR #97, opened 2020-03-23
Update default ack strategy when dealing with Message

cescoffier:
When consuming/producing Messages, the default acknowledgment should be MANUAL instead of PRE_PROCESSING.

When using messages, you should have full control. You can still set the strategy to `PRE` or `POST`. It was creating a non-homogeneous experience. So, with this change, the experience is homogeneous.

### DIFF
--- a/spec/src/main/asciidoc/architecture.asciidoc
+++ b/spec/src/main/asciidoc/architecture.asciidoc
@@ -718,7 +718,7 @@ The following table indicates the defaults and supported acknowledgement for eac
 @Incoming("channel")
 Subscriber<Message<I>> method()
 ----
-| Post-Processing	
+| Manual
 | None, Pre-Processing, Post-Processing (when the `onNext` method returns), Manual
 
 |
@@ -736,7 +736,7 @@ Subscriber<I> method()
 @Incoming("channel") 
 SubscriberBuilder<Message<I>, Void> method()
 ----
-| Post-Processing	
+| Manual
 | None, Pre-Processing, Post-Processing (when the `onNext` method returns), Manual
 
 
@@ -783,7 +783,7 @@ CompletionStage<?> method(I payload)
 @Outgoing("out")
 Processor<Message<I>, Message<O>> method()
 ----
-| Pre-Processing
+| Manual
 | None, Pre-Processing, Manual
 
 | 
@@ -804,7 +804,7 @@ Post-Processing can be optionally supported by implementations, however it requi
 @Outgoing("out") 
 ProcessorBuilder<Message<I>, Message<O>> method();
 ----
-| Pre-Processing
+| Manual
 | None, Pre-Processing, Manual
 
 |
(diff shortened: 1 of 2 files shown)


Azquelt:
I think you also want to make the the same change for:
* `Subscriber<Message<I>> method()`
* `SubscriberBuilder<Message<I>, Void> method()`
* `Processor<Message<I>, Message<O>> method()`
* `ProcessorBuilder<Message<I>, Message<O>> method()`
* `Publisher<Message<O>> method(Message<I> msg)`
* `PublisherBuilder<Message<O>> method(Message<I> msg)`

Azquelt:
This is a breaking change, so it shouldn't be attached to the 1.0.1 milestone.

If we still want to do a 1.0.1 release with non-breaking changes, we probably want to create a branch for that too.

Azquelt:
Does this mean we end up with three distict sets of signatures:
* If you're dealing with `Messages` directly, you always get manual
* Otherwise, if you're taking direct control over the processing by returning a `Publisher` or `Subscriber` at application startup, you get pre-processing
* Otherwise, you get post-processing

That would certainly make explaining the default acknowledgement strategy much easier.

cescoffier:
> Does this mean we end up with three distict sets of signatures:
>
> If you're dealing with Messages directly, you always get manual
> Otherwise, if you're taking direct control over the processing by returning a Publisher or Subscriber > at application startup, you get pre-processing
> Otherwise, you get post-processing
> That would certainly make explaining the default acknowledgement strategy much easier.

Exactly. Right now, we get users lost because it's hard to deduce the default ack policy. 
I even recommend always indicate the strategy they want - even if ti's the default one. It's not a great experience.

cescoffier:
@Azquelt I've done the requested change. You can't believe how much it simplifies the implementation code (and the mental model).

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1025,14 +1025,13 @@ public O process(I payload) {
...
 
-The `MANUAL` strategy indicates that the acknowledgment is managed by the user code. 
+The `MANUAL` strategy indicates that the acknowledgment is managed by the user code.
 The following snippet is particularly useful for processing messages that are also being sent to a destination, as the implementation must not invoke `ack` until after the outgoing message has been sent to the destination:
 
 [source, java]
 ----
 @Incoming("in")
 @Outgoing("out")
-@Acknowledgment(Acknowledgment.Strategy.MANUAL)
```
Emily-Jiang: I think this is better to change to PreProcessing to demonstrate how to set the undefault ack mode.

Emily-Jiang:
Do we have test for this?

cescoffier:
No we didn’t have any tests before.

Basic acknowledgement is tested, but not the default strategies.

spec/src/main/asciidoc/architecture.asciidoc
cescoffier: So, I decided to add one more example explaining the `PRE_PROCESSING` strategy with Messages and re-add this line with a note that it's the default strategy.

Emily-Jiang: ok

cescoffier:
@Azquelt Are you OK with the changes?

PR #106, opened 2020-07-24
Blocking IO = synchronous

nakumgaurav:
Aren't blocking transformation synchronous? I guess writing "asynchronous" must be a typo.

### DIFF
--- a/spec/src/main/asciidoc/architecture.asciidoc
+++ b/spec/src/main/asciidoc/architecture.asciidoc
@@ -239,7 +239,7 @@ It's important to remember that users must not call these methods directly.
 They are invoked by the Reactive Messaging implementation following the Reactive Streams protocol.
 
 Also the method must be implemented in a non-blocking fashion.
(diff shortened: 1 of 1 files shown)


genie-microprofile:
Can one of the admins verify this patch?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -239,7 +239,7 @@ It's important to remember that users must not call these methods directly.
 They are invoked by the Reactive Messaging implementation following the Reactive Streams protocol.
 
 Also the method must be implemented in a non-blocking fashion.
-For blocking transformations, asynchronous variants can be used.
+For blocking transformations, synchronous variants can be used.
```
danielkec, line 242: I believe the meaning is that you should execute blocking operations asynchronously, eg. don't ever block the reactive methods. Changing it to `synchronous` makes the sentence untrue.

PR #114, opened 2020-08-24
Dependency fixes

### LINKED ISSUE #75: Update to Jakarta EE8 APIs for MP 4.0 in Jun 2020

### LINKED ISSUE #88: OSGi dependency on API should be "provided"

### LINKED ISSUE #89: Spec dependencies marked as "provided"
Though I think most of the dependencies are marked as "provided" in the root `dependencyManagement` section, it's beneficial to have any Jakarta or MP API dependency marked as `provided` in /api and /tck to ensure they're not included by omission and to make it clear from looking at the child pom what the scope is.

Azquelt:
* Change EE, MP and OSGi dependencies to `provided` scope
  * Be consistent with where we set scope and version (scope in the child pom, versions in the parent pom)
* Update to Jakarta EE 8 APIs

Fixes #75
Fixes #88
Fixes #89

This may require changes to user's builds as previously including the Reactive Messaging API would pull in the MP Config and MP Reactive Streams Operators APIs as transitive dependencies but now it won't.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -39,19 +39,22 @@
         <dependency>
             <groupId>org.osgi</groupId>
             <artifactId>org.osgi.annotation.versioning</artifactId>
-            <version>1.1.0</version>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.eclipse.microprofile.reactive-streams-operators</groupId>
             <artifactId>microprofile-reactive-streams-operators-api</artifactId>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.eclipse.microprofile.config</groupId>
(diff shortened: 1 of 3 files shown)


Azquelt:
I'm not 100% sure that using `provided` scope for dependencies from the API to other MP or EE APIs is the best thing to do, but it does seem to be the MicroProfile policy.

I've made the dependency declarations more consistent, with the `version` always being specified in the parent and the `scope` always being specified in the child pom. (Not totally sure of that last part, but it's what was requested in #89)

cescoffier:
@Emily-Jiang can you check that it's what's intended. I have the same concern as Andrew.

Emily-Jiang (approved):

PR #115, opened 2020-09-23
KafkaMessage is deprecated use KafkaRecord instead

agoncal:
`io.smallrye.reactive.messaging.kafka.KafkaMessage` is marked as deprecated linking to `io.smallrye.reactive.messaging.kafka.KafkaRecord`

### DIFF
--- a/spec/src/main/asciidoc/architecture.asciidoc
+++ b/spec/src/main/asciidoc/architecture.asciidoc
@@ -65,7 +65,7 @@ Reactive Messaging application components are addressable recipients which await
 
 Messages are represented by the `org.eclipse.microprofile.reactive.messaging.Message` class. 
 This interface is intentionally kept minimal. The aim is that _connectors_ will provide their own implementations with additional metadata that is relevant to that connector.
-For instance, a `KafkaMessage` would provide access to the _topic_ and _partition_.
+For instance, a `KafkaRecord` would provide access to the _topic_ and _partition_.
 
 The `org.eclipse.microprofile.reactive.messaging.Message#getPayload` method retrieves the wrapped payload. 
 The `org.eclipse.microprofile.reactive.messaging.Message#ack` method acknowledges the message.


genie-microprofile:
Can one of the admins verify this patch?

Azquelt:
I think that `KafkaMessage` in this part of the spec is given as a hypothetical connector provided class, rather than a reference to any specific class in a particular implementation, so I'm not sure it's necessary to change it.

In addition, I think that leaving the name as `KafkaMessage` makes it more obvious that it's a subclass of `Message`

agoncal:
I should have been more precise in the issue, sorry. There are several references of `KafkaMessage` in the documentation. So it's a bit more work than just renaming `KafkaMessage` to `KafkaRecord`. `KafkaMessage` is actually used in several pieces of code, which means changing the all the code to `KafkaRecord`.

@Azquelt if `KafkaMessage` refers to a "hypothetical connector provided class, rather than a reference to any specific class" then why not renaming "`KafkaMessage`" with just "Kafka Message" or "_Kafka Message_" (in italic). `KafkaMessage` reads as code, don't you think?

Azquelt:
> why not renaming "`KafkaMessage`" with just "Kafka Message" or "_Kafka Message_" (in italic). `KafkaMessage` reads as code, don't you think?

Yes, it's an example class name, so it's formatted as a class name and used in the code example.

Certainly the wording could be improved so that it's clearer that `KafkaMessage` is given as an example of a class which a connector _could_ provide, rather than a class which implementers of the spec are required to provide or which users can expect to exist in all implementations.

danielkec:
The spec wording makes it quite clear that it is an example:
```
However, the user may decide to consume a specific subclass of Message (e.g. KafkaMessage in the following example)
```

Emily-Jiang:
Maybe we edit the spec to say `MyMessage` to make it clear.

PR #121, opened 2021-02-09
Negative Acknowledgement

cescoffier:
Add support for negative acknowledgement

* Add `nack` method to the `Message` interface
* Explain `ack` / `nack` on the `Emitter` javadoc
* Extend specification to cover negative acknowledgement
* Extend TCK to verify acknowledgement and negative-acknowledgement

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
+++ b/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
@@ -20,6 +20,7 @@ package org.eclipse.microprofile.reactive.messaging;
 
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
 import java.util.function.Supplier;
 
 /**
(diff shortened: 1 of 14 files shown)


Emily-Jiang:
@cescoffier can you raise an issue to explain the use case for `nack` so that it can be referenced in the release note?

cescoffier:
@Emily-Jiang forgot to add the linked issue, it was already created by @Azquelt.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
```diff
@@ -66,6 +67,70 @@ public T getPayload() {
...
+    }
+
+    /**
+     * Creates a new instance of {@link Message} with the given acknowledgement supplier.
+     * The payload, metadata and nack function are taken from the current {@link Message}.
+     *
+     * @param supplier the acknowledgement supplier
+     * @return the new instance of {@link Message}
+     */
+    default Message<T> withAck(Supplier<CompletionStage<Void>> supplier) {
```
Emily-Jiang, line 119: To be consistent with the following method, should `supplier` be updated to `ack`, to increase the readability?

cescoffier: Do you mean the name of the parameter? If so, yes, not opinion.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
```diff
@@ -81,6 +146,42 @@ public T getPayload() {
...
+    /**
+     * @return the function used to retrieve the negative-acknowledgement asynchronous function.
+     */
+    default Function<Throwable, CompletionStage<Void>> getNack() {
+        return reason -> CompletableFuture.completedFuture(null);
+    }
+
+    /**
+     * Acknowledge negatively this message.
+     * <code>nack</code> is used to indicate that the processing of a message failed. The reason is passed as parameter.
```
Emily-Jiang, line 165:
```suggestion
     * <code>nack</code> is used to indicate that the processing of a message failed with the reason passed as the parameter.
```

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
Emily-Jiang, line 119:
```suggestion
    default Message<T> withAck(Supplier<CompletionStage<Void>> ack) {
```

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1062,6 +1085,84 @@ public Message<O> process(Message<I> msg) {
...
+    if (s.equalsIgnoreCase("b")) {
+        // we cannot fail, we must nack explicitly.
+        m.nack(new IllegalArgumentException("b"));  // <2>
+        return null;
+    }
+    return m.withPayload(s.toUpperCase());
+}
+----
+1. `MANUAL` is the default strategy for this signature
+2. throwing an exception would not _nack_ the message, as it may have already been explicetely _acked_ or _nacked_, before the exception happens, so the user must call the `nack` method
```
Emily-Jiang, line 1161:
```suggestion
2. throwing an exception would not _nack_ the message, as it may have already been explicitly _acked_ or _nacked_. Before the exception happens, the user must call the `nack` method.
```

Emily-Jiang (approved):
LGTM!

PR #123, opened 2021-02-11
Extend @Acknowledgement javadoc to list the supported strategies and the default one for each signature

### LINKED ISSUE #111: Missing @Acknowledgement strategies from Javadoc
https://openliberty.io/docs/20.0.0.8/reference/javadoc/microprofile-3.3-javadoc.html#package=org/eclipse/microprofile/reactive/messaging/package-frame.html&class=org/eclipse/microprofile/reactive/messaging/Acknowledgment.html

In the Javadoc it gives the supported strategies for some common use cases, which I like - it's a great idea to call them out up-front rather than making the reader go through them all to work out which is most useful.

However, it then refers the reader to the specification for the full list, which I don't think it should do. The specification is aimed at developers implementing MP Reactive Messaging, not at developers using MP Reactive Messaging in their own apps; the Javadoc or product docs should provide that information.

I think it's appropriate for this info to be in the `@Acknowledgement` annotation's javadocs (as some of it is now) because it's part of what a developer needs to know to use the annotation. But all the info (strategies) needs to be there - maybe a second list showing "less common use cases" or something?

cescoffier:
Fix #111

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Acknowledgment.java
+++ b/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Acknowledgment.java
@@ -24,9 +24,19 @@ import java.lang.annotation.RetentionPolicy;
 /**
  * Configure the acknowledgement policy for the given {@code @Incoming}.
  *
+ * Reactive Messaging proposes four acknowledgement strategies:
+ * <ul>
+ *     <li><code>MANUAL</code>: the acknowledgement (positive or negative) is up to the user. This is the default strategy
+ *     for methods ingesting or producing {@link Message}.
+ *     </li>
+ *     <li><code>POST_PROCESSING</code>: acknowledges the incoming message once the produced message is acknowledged.
+ *     This is the default strategy for methods ingesting or producing single payloads.</li>
+ *     <li><code>PRE_PROCESSING</code>: acknowledges the incoming messages before calling the method.</li>
+ *     <li><code>NONE</code>: do not apply any acknowledgement.</li>
+ * </ul>
+ *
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/eclipse/microprofile/reactive/messaging/Acknowledgment.java
```diff
@@ -37,6 +47,280 @@
...
+ * </div></div></td>
+ * <td style="border: 1px solid #999; padding: 0.5rem; text-align: left;"><p>Post-Processing</p></td>
+ * <td style="border: 1px solid #999; padding: 0.5rem; text-align: left;"><p>None, Pre-Processing, Post-Processing (when
+ * the returned <code>CompletionStage</code> is completed)</p></td>
+ * </tr>
+ * <tr>
+ * <td style="border: 1px solid #999; padding: 0.5rem; text-align: left;"><div class="content"><div class="listingblock">
+ * <div class="content">
+ * <pre><code>@Incoming("in")
+ * <span style="color:#007">@Outgoing"out")
```
Emily-Jiang, line 142: @Outgoing"out") =>
@Outgoing("out")

You need to do a global replacement as there are many occurrence.

cescoffier: Damn, I fixed it for @Incoming, but not for @Outgoing.

cescoffier: Fixed.

Emily-Jiang (approved):

PR #124, opened 2021-02-11
Add @ConnectorAttribute

cescoffier:
Defines the @ConnectorAttribute and @ConnectorAttributes annotations

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/ConnectorAttribute.java
+++ b/api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/ConnectorAttribute.java
@@ -1,0 +1,127 @@
+/**
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.reactive.messaging.spi;
+
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.TYPE;
+
+/**
+ * Connector implementor can use this annotation to document attributes supported by their connector.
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/ConnectorAttribute.java
```diff
@@ -0,0 +1,127 @@
+/**
+ * Copyright (c) 2018, 2019 Contributors to the Eclipse Foundation
```
Emily-Jiang, line 2: 2021

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/ConnectorAttributes.java
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright (c) 2018, 2019 Contributors to the Eclipse Foundation
```
Emily-Jiang, line 2: 2021

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/ConnectorAttribute.java
cescoffier: oups... Fixed

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/ConnectorAttributes.java
cescoffier: fixed

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1295,6 +1295,31 @@ The configuration passed to the `IncomingConnectorFactory` and `OutgoingConnecto
 * `channel-name` attribute indicating the name of the channel being configured,
 * `connector` attribute must match the name given to the `@Connector` qualifier.
 
+==== Connector attribute
+
+To help tools (IDEs, documentation generator) to extract the configuration of each connector, the specification recommends the usage of the `org.eclipse.microprofile.reactive.messaging.spi.ConnectorAttribute` annotation.
```
Emily-Jiang, line 1300: I am wondering whether we could add an optional TCK. What do you think?

cescoffier: I don't believe we can, what would it be? 
The implementations do not do anything with these attributes.

Emily-Jiang: okay, I have not thought of a solution as yet, but I thought you might know. If you don't believe we could, it is okay.

Emily-Jiang: To help tools (IDEs, documentation generator) to extract the configuration of each connector, the specification recommends the usage of the `org.eclipse.microprofile.reactive.messaging.spi.ConnectorAttribute` annotation. =>

To help tools (IDEs, documentation generator) to extract the configuration of each connector, the specification provides the `org.eclipse.microprofile.reactive.messaging.spi.ConnectorAttribute` annotation for implementations to create a good ecosystem with the tools. However, the support for this SPI is not mandatory.

cescoffier: Fixed.

Emily-Jiang (approved):

PR #130, opened 2021-02-24
Health support

### LINKED ISSUE #47: Health support

danielkec:
Fixes #47

Signed-off-by: Daniel Kec <daniel.kec@oracle.com>

### DIFF
--- a/spec/src/main/asciidoc/architecture.asciidoc
+++ b/spec/src/main/asciidoc/architecture.asciidoc
@@ -1331,6 +1331,27 @@ The connector is responsible for the acknowledgment (positive or negative) of th
 * An outgoing connector must acknowledge the incoming `org.eclipse.microprofile.reactive.messaging.Message` once it has successfully dispatched the message.
 * An outgoing connector must acknowledge negatively the incoming `org.eclipse.microprofile.reactive.messaging.Message` if it cannot be dispatched.
 
+== Health
+
+When MicroProfile Reactive Messaging is used in an environment where MicroProfile Health is enabled, implicit readiness and liveness checks are produced.
+
+=== Readiness
+Readiness check `messaging` gets UP only when subscribers of all messaging channels are subscribed to theirs respective publishers and `onSubscribe` signal is detected on all channels.
+
+=== Liveness
+
+Liveness check `messaging` gets DOWN only when `cancel` or `onError` signals are detected on any of the messaging channels.
+
+=== Excluding channels
+
+For channels which are not desired to be checked for liveness or readiness, exclusion is configurable with `mp.health.live.exclude-channel` or `mp.health.live.exclude-channel` config keys.
+
+[source, properties]
+----
+mp.health.live.exclude-channel=<channel-name>
+mp.health.ready.exclude-channel=<channel-name>
+----
+
 == Metrics
 
 When MicroProfile Reactive Messaging is used in an environment where MicroProfile Metrics is enabled, the Reactive Messaging implementation automatically produces metrics.
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -34,6 +34,7 @@
     <properties>
         <arquillian.version>1.4.1.Final</arquillian.version>
         <shrinkwrap-api.version>1.2.6</shrinkwrap-api.version>
+        <version.glassfish.json>1.1.6</version.glassfish.json>
     </properties>
 
     <dependencyManagement>
@@ -70,19 +71,19 @@
             <artifactId>microprofile-reactive-streams-operators-api</artifactId>
             <scope>provided</scope>
         </dependency>
-        
+
         <dependency>
             <groupId>org.eclipse.microprofile.config</groupId>
             <artifactId>microprofile-config-api</artifactId>
             <scope>provided</scope>
         </dependency>
-        
+
         <dependency>
             <groupId>jakarta.enterprise</groupId>
             <artifactId>jakarta.enterprise.cdi-api</artifactId>
             <scope>provided</scope>
         </dependency>
-        
+
         <dependency>
             <groupId>jakarta.annotation</groupId>
             <artifactId>jakarta.annotation-api</artifactId>
@@ -118,6 +119,19 @@
         <dependency>
             <groupId>org.awaitility</groupId>
             <artifactId>awaitility</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>jakarta.json</groupId>
+            <artifactId>jakarta.json-api</artifactId>
+            <version>${version.glassfish.json}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>jakarta.json</artifactId>
+            <version>${version.glassfish.json}</version>
         </dependency>
 
         <dependency>
--- a/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/health/HealthAssertions.java
+++ b/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/health/HealthAssertions.java
@@ -1,0 +1,103 @@
+/**
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.reactive.messaging.tck.health;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import javax.json.Json;
+import javax.json.JsonArray;
+import javax.json.JsonObject;
+import javax.json.JsonReader;
+import javax.json.JsonValue;
+
+import static org.junit.Assert.assertEquals;
+
+class HealthAssertions {
+
+    private int responseCode;
+    private JsonObject jsonResponse;
+
+    static HealthAssertions create(String url) {
+        try {
+            HttpURLConnection con = (HttpURLConnection) new URL(url).openConnection();
+            con.setRequestMethod("GET");
+            con.setConnectTimeout(5000);
+            con.setReadTimeout(5000);
+
+            con.connect();
+            HealthAssertions healthAssertions = new HealthAssertions();
+            healthAssertions.responseCode = con.getResponseCode();
+
+            InputStream is = con.getErrorStream();
+            if(is == null){
+                is = con.getInputStream();
+            }
+            JsonReader jsonReader = Json.createReader(is);
+            healthAssertions.jsonResponse = jsonReader.readObject();
+            is.close();
+            con.disconnect();
+            return healthAssertions;
+        }
+        catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    HealthAssertions assertResponseCode(int expected){
+        assertEquals(expected, this.responseCode);
+        return this;
+    }
+
+    HealthAssertions assertResponseCodeUp(){
+        assertResponseCode(200);
+        return this;
+    }
+
+    HealthAssertions assertResponseCodeDown(){
+        assertResponseCode(503);
+        return this;
+    }
+
+    HealthAssertions assertMessagingStatus(String expectedStatus){
+        JsonArray checks = this.jsonResponse.getJsonArray("checks");
+        List<JsonObject> messagingObjects = checks.stream()
+            .map(JsonValue::asJsonObject)
+            .filter(o -> o.containsKey("name") && "messaging".equals(o.getString("name")))
+            .collect(Collectors.toList());
+        assertEquals(1, messagingObjects.size());
+        JsonObject messagingObject = messagingObjects.get(0);
+        assertEquals(expectedStatus, messagingObject.getString("status"));
+        return this;
+    }
+
+    HealthAssertions assertMessagingStatusUp(){
+        assertMessagingStatus("UP");
+        return this;
+    }
+
+    HealthAssertions assertMessagingStatusDown(){
+        assertMessagingStatus("DOWN");
+        return this;
+    }
+}
--- a/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/health/HealthCancelledConnectorInChannelTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/health/HealthCancelledConnectorInChannelTest.java
@@ -1,0 +1,127 @@
+/**
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.reactive.messaging.tck.health;
+
+import javax.inject.Inject;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.junit.InSequence;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Test;
+import org.junit.runner.RunWith;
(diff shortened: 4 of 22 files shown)


spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1331,6 +1331,27 @@ The connector is responsible for the acknowledgment (positive or negative) of th
...
+=== Readiness
+Readiness check `messaging` gets UP only when subscribers of all messaging channels are subscribed to theirs respective publishers and `onSubscribe` signal is detected on all channels.
+
+=== Liveness
+
+Liveness check `messaging` gets DOWN only when `cancel` or `onError` signals are detected on any of the messaging channels.
+
+=== Excluding channels
+
+For channels which are not desired to be checked for liveness or readiness, exclusion is configurable with `mp.health.live.exclude-channel` or `mp.health.live.exclude-channel` config keys.
```
Azquelt, line 1347: I assume one of these should say `mp.health.ready.exclude-channel` ?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1331,6 +1331,27 @@ The connector is responsible for the acknowledgment (positive or negative) of th
 * An outgoing connector must acknowledge the incoming `org.eclipse.microprofile.reactive.messaging.Message` once it has successfully dispatched the message.
 * An outgoing connector must acknowledge negatively the incoming `org.eclipse.microprofile.reactive.messaging.Message` if it cannot be dispatched.
 
+== Health
+
+When MicroProfile Reactive Messaging is used in an environment where MicroProfile Health is enabled, implicit readiness and liveness checks are produced.
+
+=== Readiness
```
tomas-langer, line 1338: Please reword - such as "gets UP" should be reports `UP`.
Maybe a bit more precise:
- Readiness reports `DOWN` when messaging starts
- Once all subscribers.... the state changes to `UP`

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1331,6 +1331,27 @@ The connector is responsible for the acknowledgment (positive or negative) of th
...
+Readiness check `messaging` gets UP only when subscribers of all messaging channels are subscribed to theirs respective publishers and `onSubscribe` signal is detected on all channels.
+
+=== Liveness
+
+Liveness check `messaging` gets DOWN only when `cancel` or `onError` signals are detected on any of the messaging channels.
```
tomas-langer, line 1343: Similar to Readiness - define if it starts in `UP` or `DOWN` state, then define states that trigger change

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1331,6 +1331,27 @@ The connector is responsible for the acknowledgment (positive or negative) of th
...
+
+=== Excluding channels
+
+For channels which are not desired to be checked for liveness or readiness, exclusion is configurable with `mp.health.live.exclude-channel` or `mp.health.live.exclude-channel` config keys.
+
+[source, properties]
+----
+mp.health.live.exclude-channel=<channel-name>
```
tomas-langer, line 1351: You should not use Health namespace here, as that is owned by the Health specification.
Also explicitly mark this as an array (as `channel` singular does not show it is possible to define more than one, nor is it mentioned in text).
`mp.messaging.health.live.exclude=channel1,channel2` or similar

tck/pom.xml
```diff
@@ -70,19 +71,19 @@
             <artifactId>microprofile-reactive-streams-operators-api</artifactId>
             <scope>provided</scope>
         </dependency>
-        
+
```
tomas-langer, line 74: Remove the newlines added

tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/health/HealthCancelledConnectorInChannelTest.java
```diff
@@ -0,0 +1,127 @@
...
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.junit.InSequence;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(Arquillian.class)
+public class HealthCancelledConnectorInChannelTest extends HealthBase {
```
tomas-langer, line 32: Each test should have javadoc.
Test class: which area of the spec is covered
Test method: what explicit statement from spec it tests

tomas-langer (changes requested):

spec/src/main/asciidoc/architecture.asciidoc
Emily-Jiang, line 1351: We should not prefix with mp.health. All o fthe configuration introduced in this spec will be prefixed with mp.messaging.

Emily-Jiang: I guess this should be plurual as you might want to exclude multiple channels with comma separated.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1331,6 +1331,27 @@ The connector is responsible for the acknowledgment (positive or negative) of th
 * An outgoing connector must acknowledge the incoming `org.eclipse.microprofile.reactive.messaging.Message` once it has successfully dispatched the message.
 * An outgoing connector must acknowledge negatively the incoming `org.eclipse.microprofile.reactive.messaging.Message` if it cannot be dispatched.
 
+== Health
+
+When MicroProfile Reactive Messaging is used in an environment where MicroProfile Health is enabled, implicit readiness and liveness checks are produced.
+
+=== Readiness
+Readiness check `messaging` gets UP only when subscribers of all messaging channels are subscribed to theirs respective publishers and `onSubscribe` signal is detected on all channels.
```
Emily-Jiang, line 1339: What do you mean by saying check `messaging`?

spec/src/main/asciidoc/architecture.asciidoc
Emily-Jiang: What should happen if a message gets `nack`ed?

spec/src/main/asciidoc/architecture.asciidoc
danielkec: thx fixed

spec/src/main/asciidoc/architecture.asciidoc
danielkec: thx reworded

spec/src/main/asciidoc/architecture.asciidoc
danielkec: I don't believe acknowledgement should be related to the ability of the channel to send messages as messaging itself doesn't leverage acking/nacking anyhow. It should be responsibility of the publishers and subscribers(connectors) as that requirement can change case by case. In other words such checks should be present in the connectors or business code if or when needed.

spec/src/main/asciidoc/architecture.asciidoc
danielkec: Yes you are totally right, switched to `mp.messaging.health` context

tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/health/HealthCancelledConnectorInChannelTest.java
danielkec: Added javadoc with detailed step by step description of the test scenario

spec/src/main/asciidoc/architecture.asciidoc
danielkec: You are right, moved to `mp.messaging.health` context and fixed the example

spec/src/main/asciidoc/architecture.asciidoc
danielkec: Its the name of the check itself, added example and reworded

tck/pom.xml
danielkec: Its just removed trailing whitespace and I am literally unable to revert it https://youtrack.jetbrains.com/issue/IDEA-244932

danielkec: Fixed

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1331,6 +1331,44 @@ The connector is responsible for the acknowledgment (positive or negative) of th
 * An outgoing connector must acknowledge the incoming `org.eclipse.microprofile.reactive.messaging.Message` once it has successfully dispatched the message.
 * An outgoing connector must acknowledge negatively the incoming `org.eclipse.microprofile.reactive.messaging.Message` if it cannot be dispatched.
 
+== Health
+
+When MicroProfile Reactive Messaging is used in an environment where MicroProfile Health is enabled, implicit readiness
```
cescoffier, line 1336: readiness might be misleading. Do you mean "ready to get traffic" or "started successfully"?  Kubernetes has readiness and startup checks, and readiness is focusing on traffic, which most of the time is ignored by messaging protocol (as the probe are used for HTTP request routing).

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1331,6 +1331,44 @@ The connector is responsible for the acknowledgment (positive or negative) of th
...
+and liveness checks `messaging` are produced.
```
cescoffier, line 1337: `messaging` might not be explicit enough, or conflict with user checks. What about `MicroProfile Reactive Messaging`?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1331,6 +1331,44 @@ The connector is responsible for the acknowledgment (positive or negative) of th
...
+  "status": "UP",
+  "checks": [
+    {
+      "name": "messaging",
+      "status": "UP"
+    }
+  ]
+}
+----
+
```
cescoffier, line 1351: In the case, we don't pass the checks, what extra data should we provide? None won't be very useful (even if these checks tend to be consumed by machines ignoring the extra data). Should we list the unsatisfied channels? Is it implementation specific?

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1331,6 +1331,44 @@ The connector is responsible for the acknowledgment (positive or negative) of th
...
+=== Readiness
+State of the readiness check `messaging` is DOWN until subscribers of all messaging channels are subscribed to their
+respective publishers and `onSubscribe` signal is detected on all channels.
+
+=== Liveness
+
+State of the liveness check `messaging` is UP until `cancel` or `onError` signals are detected on any of the messaging
+channels.
```
cescoffier, line 1359: Not sure about cancellations. We have seen applications using cancellations as part of their regular lifecycle. 
Failure definitely.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1331,6 +1331,44 @@ The connector is responsible for the acknowledgment (positive or negative) of th
...
+
```
cescoffier, line 1360: Wondering about "all channels" while it seems to only require checking the _final_ subscribers (subscriber methods, outgoing channels).

Also, about `@Channel` injection (unmanaged stream). It might be hard to track, as in this case, the subscription is delegated to the users. So, all the used channels may not receive their subscription, because the user has not subscribed yet (and it does not mean it's not ready). 

Maybe it should focus on:

* subscriber methods:
  * for readiness, must have received a subscription, 
  * for liveness, a received failure would trigger the liveness check to be DOWN
* outgoing connector 
   * for readiness, must have received a subscription, 
   * for liveness, a received failure would trigger the liveness check to be DOWN, an unrecoverable (like serialization) failure in the connector logic would trigger the liveness check to be DOWN.
*  incoming connector
   *  for liveness, any unrecoverable failure would set the liveness check to DOWN
   * for readiness, it can be tricky, as we may not have a downstream subscription yet (because of unmanaged streams) (we are actually seen reports about that in Quarkus)
   
Intermediate channels can recover from failures, and implement retry logic, so it should not be reported as a DOWN.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -1331,6 +1331,44 @@ The connector is responsible for the acknowledgment (positive or negative) of th
...
+=== Excluding channels
+
+For channels which are not desired to be checked for liveness or readiness, exclusion is configurable with
+`mp.messaging.health.live.exclude` or `mp.messaging.health.ready.exclude` config keys.
+
+[source, properties]
+----
+mp.messaging.health.live.exclude=<channel-name>,<channel-name>
+mp.messaging.health.ready.exclude=<channel-name>,<channel-name>
```
cescoffier, line 1369: If we follow the logic I described above, this would only make sense for subscriber methods. For connectors, it would be better to have a specific attribute and keep the connector configuration co-localized.

tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/health/HealthAssertions.java
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Connect health endpoint and download response to be able to assert it later.
+     *
+     * @param url of the health check endpoint
+     * @return new instance of the health assertions
+     */
+    static HealthAssertions create(String url) {
+        try {
+            HttpURLConnection con = (HttpURLConnection) new URL(url).openConnection();
```
cescoffier, line 52: Do we have to use HTTP to verify the checks? Doesn't Microprofile Health expose an API (asking the question as I never checked).

cescoffier (changes requested):
I did a quick first pass.

In addition to the comments, I made in the code, the TCK does not seem to cover Emitter, unmanaged channels, and nacked messages. 

The TCK relies on HTTP, which is a bit annoying, as it makes running the TCK much more framework-specific. Typically, I would need to use Quarkus (and so create a cycle between out implementation and Quarkus), instead of just Weld.

spec/src/main/asciidoc/architecture.asciidoc
danielkec: We have also seen use-cases with failures as part of its regular lifecycle (onErrorResumeWith, ...), but as I see it channel is a simple processor which is not mutating the stream(except for un/wrapping), so any `cancel` or `onError` signal passing thru channel renders it DOWN for good

tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/health/HealthCancelledConnectorInChannelTest.java
```diff
@@ -0,0 +1,145 @@
...
+ * <li>8. Send cancel signal to one of the channels</li>
+ * <li>9. Assert {@code /health/live} is DOWN as cancel signal has been issued on one channel</li>
+ * <li>10. Assert {@code /health/ready} is still UP</li>
+ * <li>11. Send cancel signal to all the other channels</li>
+ * <li>12. Assert {@code /health/live} is still DOWN</li>
+ * <li>13. Assert {@code /health/ready} is still UP</li>
+ * </ul>
+ */
+@RunWith(Arquillian.class)
+public class HealthCancelledConnectorInChannelTest extends HealthBase {
```
Emily-Jiang, line 50: The test should only run if MP Health is present. One of the solutions is to load Readiness api to check whther MP Health is enabled or not. This comment apply to all of the tests.

spec/src/main/asciidoc/architecture.asciidoc
danielkec: Problem with channels where retries/re-subscriptions are expected should be solvable with exclusions(inclusions?)

spec/src/main/asciidoc/architecture.asciidoc
danielkec: This would be quite a different angle than the one I was trying to take. 

There are not much parts of reactive pipelines we are actually managing in the MP Messaging impls and those are the channels. Channels are basically processors which are not mutating the stream (except for un/wrapping) so if `cancel` or 'onError' signal passes thru it, it can be considered as failed. Any recovery made by unmanaged streams is just re-routing the stream thru different chain of operators or retry/resubscribe.

I am not sure that the health of connectors, methods and unmanaged streams should be in the scope of MP Messaging implicit health check. I would count those as business code where health and readiness can be monitored with standard  MP Health API.

Also implicit checks should be consistent, if we differentiate its behavior by its pub/sub kinds, we risk user confusion. Again custom check with exclusion/inclusion can help in such cases.

spec/src/main/asciidoc/architecture.asciidoc
danielkec: I was playing with that idea but couldn't come with any sufficient reason for requiring channel listing in the checks

spec/src/main/asciidoc/architecture.asciidoc
danielkec: I ve been thinking about use-cases of producing/consuming RS channels with websocket or SSE

spec/src/main/asciidoc/architecture.asciidoc
danielkec: You are right, I just can't find any definitive convention for naming health checks, maybe `mp-reactive-messaging` would make it more machine readable?

tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/health/HealthAssertions.java
danielkec: That would be great, I didn't found other way. MP Health TCKs use same approach

tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/health/HealthCancelledConnectorInChannelTest.java
danielkec: IMHO if the feature is required when mp health is present, mp health needs to be required by TCKs. Other way around feature would be effectively optional.

Emily-Jiang: The MP RM feature does not need MP Health to be enabled. Only this integration comes into action if MP health is enabled. My point is that the tck will fail if a vendor does not support MP Health. However, MP Heath is not mandatory hence the tcks should only be activated if MP health api can be loaded in the runtime.

PR #158, opened 2023-02-21
Replace flaky assertion on DropOverflowStrategyOverflowTest

ozangunalp:
It is highly unlikely but the list of undropped payloads can contain `"999"`. We know for sure that the number of emitted messages will be less than what is trying to be sent.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/channel/overflow/DropOverflowStrategyOverflowTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/channel/overflow/DropOverflowStrategyOverflowTest.java
@@ -44,7 +44,7 @@ public class DropOverflowStrategyOverflowTest extends TckBase {
 
         bean.emitALotOfItems();
         await().until(bean::isDone);
-        assertThat(bean.output()).isNotEmpty().doesNotContain("999");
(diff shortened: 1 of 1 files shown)


kabir:
FWIW we see occasional failures in this test in some of our testsuite runs

cescoffier (approved):

Azquelt:
I'm unsure that this test is actually effective. The test waits for the async thread to have emitted all the events but doesn't wait for the returned CompletionStages to complete before checking the received items.

If there's any kind of buffering, I think it's possible that this test would pass even if items weren't dropped.

Azquelt:
That said, I do acknowledge that the test as currently written is definitely wrong and payload "999" could be delivered.

ozangunalp:
Unless we setup some timeout, we can't wait for CompletionStages. For dropped items, they will never complete.

Azquelt:
Hmm, you might be right there. I expected the message to be nack'd when it was dropped so the CompletionStage would complete exceptionally but I can't see anything in the spec that requires that.

Azquelt:
I want to do a release with other TCK fixes soon and I'd forgotten about this PR. I'm still not sure that the test is now effective, and I'm not sure if we can make it effective if we can't rely on all messages either being acked or nacked, but at least with this change it won't be broken so we should go ahead and merge it.

PR #159, opened 2023-08-29
Include TckBase.class in application archives where test class extends TckBase

abutch3r:
During running of the 3.0 TCK. It was found that for a number of the newer test classes that extended TckBase, did not include the TckBase class in the application archive would throw ClassNotFoundException TckBase when run on Open Liberty.

Use the `getBaseArchive()` to include base classes and configuration and only add the additional classes for those tests.

All the affected tests use the same ServiceLoader and manifest resource as defined in `getBaseArchive()`

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/acknowledgement/AsynchronousMessageProcessorAckTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/acknowledgement/AsynchronousMessageProcessorAckTest.java
@@ -22,7 +22,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
 
 import java.util.List;
-import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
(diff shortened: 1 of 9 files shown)


Emily-Jiang (approved):
LGTM! Thanks @abutch3r

Azquelt:
This is the same problem that we've fixed before in #56.

It looks like the tests fixed in this PR are new ones added since 1.0.

Azquelt:
In case it's not clear, this is a TCK challenge. The tests are broken and cannot be run in a remote or managed arquillian container.

Azquelt:
The build is failing with the following error:
```
Exception in thread "main" java.lang.UnsupportedClassVersionError: hudson/remoting/Launcher has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

This appears to be because the Jenkins instance has been updated to a version which requires Java 11 and the maven integration plugin [requires that the build is run with a version of Java which is at least as high as that required by Jenkins](https://plugins.jenkins.io/maven-plugin/).

I'll update the build definition to run with Java 11.

Azquelt:
@eclipse-microprofile-bot test this please

Azquelt:
The build now fails because the version of `impsort-maven-plugin` we're using does not support the latest version of maven which the build is configured to use.

We should update to the latest 2.x version of the microprofile-parent pom to update our plugins and dependencies.

Azquelt:
With #160 merged, let's try again

@eclipse-microprofile-bot test this please

Azquelt:
Builds are now passing and I've tested this change with #160 on both Open Liberty and smallrye-reactive-messaging.

